PR #5, opened 2018-07-04
Seer edf downloader

shannonclarke:
example to download data from the platform and save as an edf file, in parallel


Examples/SeerEDFDownloader.py
```diff
@@ -0,0 +1,195 @@
...
+import tkinter as tk
+import traceback
+import multiprocessing
+    
+
+def dummpyTK():
+    root = tk.Tk()
+    root.withdraw()
+    root.destroy()
+    root = None
```
robrkerr: What's this for? Just a way of making sure no dialog currently exists?

Examples/SeerEDFDownloader.py
```diff
@@ -0,0 +1,195 @@
...
+        import seerpy
+        client = seerpy.SeerConnect()
+        
+        
+        q = multiprocessing.Queue(9999)
+        procs = [multiprocessing.Process(target=dlData, args=(q,client)) for i in range(5)]
+#        for p in procs: p.daemon = False
+        for p in procs: p.start()
+        
+        study = input('Study to Download: ')
```
robrkerr: How hard would it be to make this a GUI dialog where the user selects a study from a list? Probably not urgent but it also might make your life easier instead of having to enter the name perfectly each time.

PR #10, opened 2018-10-04
Handle unprocessed segments

shannonclarke:
checks if segment is being processed before attempting to download and warns user. also adds ability to query what data has been viewed.

### DIFF
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -359,13 +359,13 @@ class SeerConnect:
                     actualChannelNames = channelNames if len(channelNames) == numChannels else ['Channel %s' % (i) for i in range(0, numChannels)]
 
                     metaData = metaData.drop_duplicates('segments.id')
+                    unprocessedSegments = metaData[metaData['segments.processed'] == 'null']
+                    if len(unprocessedSegments) > 0:
+                        print('Warning: There are ', len(unprocessedSegments), ' unprocessed segments. Consider waiting until these are finished')
                     
                     fromTime = metaData['segments.startTime'].min()
                     toTime = fromTime + metaData['segments.duration'].sum()
                     dataChunks = self.getDataChunks(studyID, channelGroupsID, fromTime, toTime)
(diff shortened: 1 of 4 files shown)


seerpy/seerpy.py
```diff
@@ -359,6 +359,12 @@ def getLinks(self, allData, threads=None):
                     actualChannelNames = channelNames if len(channelNames) == numChannels else ['Channel %s' % (i) for i in range(0, numChannels)]
 
                     metaData = metaData.drop_duplicates('segments.id')
+                    unprocessedSegments = metaData[metaData['segments.processed'].isnull()]
+                    if len(unprocessedSegments) > 0:
+                        print('Warning: There are ', len(unprocessedSegments), ' unprocessed segments. Consider waiting until these are finished')
+                    metaData = metaData[metaData['segments.processed'].notnull()]
+                    if len(metaData) == 0:
+                        continue
```
bjdoyle: If you don't use unprocessedSegments elsewhere, would it be more efficient to check the length of the processed segments against the entire length? You'd have to keep a reference to it though.

seerpy/seerpy.py
```diff
@@ -359,12 +367,6 @@ def getLinks(self, allData, threads=None):
                     actualChannelNames = channelNames if len(channelNames) == numChannels else ['Channel %s' % (i) for i in range(0, numChannels)]
 
                     metaData = metaData.drop_duplicates('segments.id')
-                    processedSegments = metaData[metaData['segments.processed'].notnull()]
-                    if len(processedSegments) < len(metaData):
-                        print('Warning: There are ', len(metaData) - len(processedSegments), ' unprocessed segments. Consider waiting until these are finished')
-                    metaData = processedSegments
-                    if len(metaData) == 0:
-                        continue
```
bjdoyle: I don't know enough about what's going on here to make an informed judgement, but I don't see any problem. Except, do you want to keep the last 2 lines checking for length 0?

shannonclarke: good idea. added back in.

PR #18, opened 2019-07-22
Remove deprecated 'party_id' GraphQL parameters and replace with url query parameter

robrkerr:
This update is for `seer-py` to continue working with the API - there was a breaking change to Seer Cloud recently that broke `seer-py`.

### DIFF
--- a/seerpy/graphql.py
+++ b/seerpy/graphql.py
@@ -141,11 +141,10 @@ def get_segment_urls_query_string(segment_ids):
         }""" % segment_ids_string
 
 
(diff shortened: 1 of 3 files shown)


bjdoyle:
The tests are broken. (They don't automatically run in build kite you have to unblock them). Let me know if you want me to look at them.

robrkerr:
> The tests are broken. (They don't automatically run in build kite you have to unblock them). Let me know if you want me to look at them.

Ah I see. Yeah I can look at that.

PR #21, opened 2019-08-21
Add label group type to label groups query

bjdoyle:
This PR adds the labelType field to the labelGroup query.

It also adds seerpy to it's own requirements.txt as an editable install so that it works properly in a virtual environment, and updates the .gitignore so virtual environment files are not included in git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,111 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
 
-*.xml
-*.iml
-/Graph*
-*/__pycache__/*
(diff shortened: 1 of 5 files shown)


bjdoyle:
I added seerpy to its own requirements mostly because of the problems people were having the other day with the downloader. They had obviously previously installed the precious version and then updated the git repository but not re-installed seerpy, so when they ran the new contest downloader it was referencing the installed old version of seerpy. This doesn't solve that problem, but does mean that if you install from the requirments.txt then the examples will work.
Turned out I needed to add a few files to the initial copy to app in the docker file for it to work on build kite, hence the changes.

PR #32, opened 2019-11-12
Optimise get_study_ids_from_names

seerwill:
Previously the entire dataset was downloaded and filtered
locally in the DataFrame. This approach runs one query per
study ID and concatenates the resulting DataFrames. This
should perform faster in most cases. Closes #28


seerpy/seerpy.py
```diff
@@ -292,8 +292,16 @@ def get_studies_dataframe(self, limit=50, search_term='', party_id=None):
     def get_study_ids_from_names_dataframe(self, study_names, party_id=None):
         if isinstance(study_names, str):
             study_names = [study_names]
-        studies = self.get_studies_dataframe(party_id=party_id)
-        return studies[studies['name'].isin(study_names)][['name', 'id']].reset_index(drop=True)
+
+        studies = pd.concat([
+            self.get_studies_dataframe(search_term=study_name, party_id=party_id) \
+                for study_name in study_names
+        ], join='outer')
```
bjdoyle: I wonder if it would be better to call get_studies() and combine the returned lists and json_nomalize() them all at once. You would be duplicating some code from get_studies_dataframe() but it might perform better. You could also refactor get_studies_dataframe() so that they both share the shared functionality.

seerwill: I have refactored to use `get_studies`, but I don't think there is enough common functionality to refactor `get_studies_dataframe()` as well. 

I've also added an additional test for querying multiple IDs as this was not covered by previous tests.

seerpy/seerpy.py
```diff
@@ -292,8 +292,16 @@ def get_studies_dataframe(self, limit=50, search_term='', party_id=None):
     def get_study_ids_from_names_dataframe(self, study_names, party_id=None):
         if isinstance(study_names, str):
             study_names = [study_names]
-        studies = self.get_studies_dataframe(party_id=party_id)
-        return studies[studies['name'].isin(study_names)][['name', 'id']].reset_index(drop=True)
+
+        studies = json_normalize([
+            x for study_name in study_names
+            for x in self.get_studies(search_term=study_name, party_id=party_id)
+        ])
```
bjdoyle: A more descriptive name than x would be good. I'd suggest study but it could actually be more than one. And studies is already taken. I can see why you opted for x, but it would still be good to be more descriptive about what it is, particularly in a multi-level list comprehension when the cognitive load is already high.

tests/test_seerpy.py
```diff
@@ -145,17 +145,49 @@ def test_existing_study_param(self, seer_auth, gql_client, unused_time_sleep):
...
+            side_effects.append({'studies': json.load(f)})
+        # this is the "no more data" response for get_studies()
+        side_effects.append({'studies': []})
+
+        with open(TEST_DATA_DIR / "studies_filtered2.json", "r") as f:
             side_effects.append({'studies': json.load(f)})
         # this is the "no more data" response for get_studies()
         side_effects.append({'studies': []})
+        # this is the "no more data" response for get_studies()
+        side_effects.append({'studies': []})
```
bjdoyle: Is this second empty response required?

PR #34, opened 2019-11-19
Pagination for diary labels

pkaroly:
This PR *elegantly* adds a limit and offset to the `get_diary_labels` so that all available labels are returned (previously this was capped to 100). 

Pair coding with @dominique-eden :)

## File changes
- Added a new example code to download and save labels from patient seizure diaries
- Included `limit` and `offset` in the `get_diary_labels_query_string` query as well as the `numberOfLabels` field (a nice sanity check to make sure you are getting all the available labels)
- Update `get_diary_labels` to loop through and request all labels for all label groups
- Also my linter seems to have made some random whitespace changes around about (sorry).

### DIFF
--- a/Examples/download_diary_seizures.py
+++ b/Examples/download_diary_seizures.py
@@ -6,12 +6,17 @@ patients = seer_client.get_patients_dataframe() #party_id can be blank to return
 for n in range(len(patients)):
 
     patient_id = patients.loc[n]['id']
-    diary_df = seer_client.get_diary_labels_dataframe(patient_id)
-    if 'labelType' in diary_df.columns and not diary_df.loc[diary_df['labelType'] == 'seizure'].empty:
-        nEvents = diary_df.loc[diary_df['labelType'] == 'seizure'].iloc[0]['numberOfLabels']
-        if nEvents >= 10:
-            events = diary_df.loc[diary_df['labelType'] == 'seizure']
-            mdict = {'data': events}
-            savename = 'Patient '  + str(n)
-            print('saving ' + savename + '...')
-            events.to_excel(savename + '.xlsx')
+    patient_email = patients.loc[n]['user.email']
+    
+    if patient_email == 'pip@seermedical.com':
+        print(patient_email)
+        diary_df = seer_client.get_diary_labels_dataframe(patient_id)
+        
+#        if 'labelType' in diary_df.columns and not diary_df.loc[diary_df['labelType'] == 'seizure'].empty:
+#            nEvents = diary_df.loc[diary_df['labelType'] == 'seizure'].iloc[0]['numberOfLabels']
+#            if nEvents >= 10:
+#                events = diary_df.loc[diary_df['labelType'] == 'seizure']
+#                mdict = {'data': events}
+#                savename = 'Patient '  + str(n)
+#                print('saving ' + savename + '...')
+#                events.to_excel(savename + '.xlsx')
--- a/Examples/download_diary_seizures.py
+++ b/Examples/download_diary_seizures.py
@@ -15,9 +15,9 @@ for patient_id in patient_ids:
     diary_df = seer_client.get_diary_labels_dataframe(patient_id)
 
 #    Limit label groups to people with an epilepsy diary (where the default labelType is seizure)
-    if diary_df is not None and \
-        'labelType' in diary_df.columns \
-            and not diary_df.loc[diary_df['labelType'] == 'seizure'].empty:
+    if (diary_df is not None
+        and 'labelType' in diary_df.columns
+            and not diary_df.loc[diary_df['labelType'] == 'seizure'].empty):
 
         n_events = diary_df.loc[diary_df['labelType']
                                 == 'seizure'].iloc[0]['numberOfLabels']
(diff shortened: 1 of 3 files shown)


seerpy/seerpy.py
```diff
@@ -447,24 +447,52 @@ def get_documents_for_studies_dataframe(self, study_ids, limit=50):
                 documents.append(document)
         return pd.DataFrame(documents)
 
-    def get_diary_labels(self, patient_id):
-        query_string = graphql.get_diary_labels_query_string(patient_id)
+    def get_diary_labels(self, patient_id, offset=0, limit=100):
+        query_string = graphql.get_diary_labels_query_string(patient_id, limit, offset)
         response = self.execute_query(query_string)['patient']['diary']['labelGroups']
```
bjdoyle: Do you want to get only a page or all pages? You can get all pages if you change the execute_query call to get_paginated_response.

bjdoyle: Oh, I see, you're checking each group below.

pkaroly: I tried using get_paginated_response but it didn't seem to fit the data format of ['patient']['diary']['labelGroups']. So I wrote the pagination loop into get_diary_labels_dataframe instead. Open to other suggestions! I was just being expedient because I wanted to download and use the data for a paper 😉

seerpy/seerpy.py
```diff
@@ -471,24 +471,52 @@ def get_documents_for_studies_dataframe(self, study_ids, limit=50):
...
```
bjdoyle: I think the way it's done in get_labels() and get_labels_dataframe() is better - both functions will return all labels and it makes the data frame version much simpler. Your case is slightly more complicated as you can have multiple labelGroups but otherwise they are very similar.

bjdoyle:
Whoops, I didn't mean to delete your comment Pip! I was just deleting mine as I realised it wasn't relevant - I didn't see your comment before I deleted!

Examples/download_diary_seizures.py
```diff
@@ -0,0 +1,28 @@
...
+for n in range(len(patients)):
+
+    patient_id = patients.loc[n]['id']
+    
+    diary_df = seer_client.get_diary_labels_dataframe(patient_id)
+    
+#    Limit label groups to people with an epilepsy diary (where the default labelType is seizure)
+    if 'labelType' in diary_df.columns and not diary_df.loc[diary_df['labelType'] == 'seizure'].empty:
+        
+        nEvents = diary_df.loc[diary_df['labelType'] == 'seizure'].iloc[0]['numberOfLabels']
```
bjdoyle: better to use snake_case than camelCase for consistency

Examples/download_diary_seizures.py
```diff
@@ -0,0 +1,28 @@
...
+
+import seerpy
+
+N_EVENTS = 10
+
+seer_client = seerpy.SeerConnect()
+
+patients = seer_client.get_patients_dataframe() #party_id can be blank to return all results
+
+for n in range(len(patients)):
```
bjdoyle: I think you're better to get the patient ids in a list and iterate that.
although you use n to name the patients, if you want to keep that then probably leave as is.

Examples/download_diary_seizures.py
```diff
@@ -0,0 +1,28 @@
...
+        
+        nEvents = diary_df.loc[diary_df['labelType'] == 'seizure'].iloc[0]['numberOfLabels']
+        
+        if nEvents >= N_EVENTS:
+#            only save the events of interest
+            events = diary_df.loc[diary_df['labelType'] == 'seizure']
+            mdict = {'data': events}
+            savename = 'Patient '  + str(n)
+            print('saving ' + savename + '...')
+            events.to_excel(savename + '.xlsx')
```
bjdoyle: not csv?

seerpy/seerpy.py
```diff
@@ -521,17 +521,46 @@ def get_documents_for_studies_dataframe(self, study_ids, limit=50):
...
         return pd.DataFrame(documents)
 
-    def get_diary_labels(self, patient_id):
-        query_string = graphql.get_diary_labels_query_string(patient_id)
-        response = self.execute_query(query_string)['patient']['diary']['labelGroups']
-        return response
+    def get_diary_labels(self, patient_id, offset=0, limit=100):
+        label_results = None
+        # set true if we need to fetch labels
+        query_flag = 1
```
bjdoyle: I think this is better as a boolean

Examples/download_diary_seizures.py
pkaroly: 🐍 🐫

Examples/download_diary_seizures.py
pkaroly: For my intended use I wanted unique but de-identified save names, because the cohort is always changing so naming them P1, P2, etc is not great for tracking people's data.

pkaroly: Oh your way is still better though. I'll change it

Examples/download_diary_seizures.py
pkaroly: Again, this is just my own use case.. Other users can always edit the example code on their own though right?

bjdoyle: Of course.

PR #40, opened 2020-01-24
get tags with label string, also add new with_tags parameter

shannonclarke:


norrishd:
Super old PR with no explanation and presumably no one left who knows what it's for; gonna close for now but obviously it can be re-opened if someone wants this.

PR #42, opened 2020-02-11
Mood survey results

seerwill:
This PR adds the queries to retrieve the results of mood surveys that the user has access to. The user must know the `surveyTemplateIds` that they wish to retrieve results for. 

Results can be obtained as either a list of survey responses (list of dictionaries) or as a flattened pandas DataFrame.


tests/test_data/mood_survey_response.json
```diff
@@ -0,0 +1,36 @@
+[
```
bjdoyle: I think this should actually be _result.json and the other 2 should be _response.json, as they are the response from the query, and this is the end result, and this is the equivalent of _results.csv.

tests/test_seerpy.py
```diff
@@ -398,3 +398,38 @@ def test_success(self, seer_auth, gql_client, unused_time_sleep):
 
         # check result
         pd.testing.assert_frame_equal(result, expected_result, check_like=True)
+
```
bjdoyle: For completeness I would add a test for no result and a test for multiple responses containing data instead of just 1.
It's possible I'm asking you to do something I didn't do 😄 but in my defence I was adding multiple tests after the fact and just trying to get a minimum of coverage. Going forward it would be good if we could test all main logic paths.

PR #43, opened 2020-02-11
Add StudyCohort queries / mutations

seerwill:
Adds mutations and queries required for getting, creating and editing StudyCohorts in the API.

Contains four methods:

1) getStudiesInStudyCohort which retrieves all of the study IDs in a given cohort. It returns a list rather than a DataFrame as only a single column is returned and the typical use case is to iterate over the study IDs

2) createStudyCohort
3/4) addStudyTo/removeStudyFromStudyCohort


seerpy/seerpy.py
```diff
@@ -838,3 +838,32 @@ def get_diary_fitbit_data(self, segments):
             data = None
 
         return data
+
+    def get_study_ids_in_study_cohort(self, study_cohort_id, page_size=200, offset=0):
```
bjdoyle: I would use limit instead of page_size to be consistent

seerpy/graphql.py
```diff
@@ -534,3 +535,79 @@ def get_diary_study_channel_groups_query_string(patient_id, from_time, to_time):
...
+                name
+                studies(limit: %0.f, offset: %0.f) {
+                    id
+                }
+            }
+        }
+    """ % (utils.quote_str(study_cohort_id), limit, offset)
+
+
+def create_study_cohort_mutation_string(name, description=None, key=None, study_ids=None):
```
bjdoyle: The mutations aren't referenced in seerpy.py - is this coming later?

seerpy/graphql.py
```diff
@@ -534,3 +535,79 @@ def get_diary_study_channel_groups_query_string(patient_id, from_time, to_time):
                 }
             }
         }""" % (patient_id, from_time, to_time)
+
+
+def get_study_ids_in_study_cohort_query_string(study_cohort_id, limit, offset):
```
bjdoyle: I haven't changed all of the existing ones, but I tend to use f-strings on new queries.
No need to change these but something to keep in mind for new ones.

seerpy/seerpy.py
seerwill: Oh yeah sorry, I discussed this with Shannon this morning. I'll fix it.

seerpy/graphql.py
bjdoyle: as discussed, ignore this

PR #45, opened 2020-02-14
Support different API endpoints

seerwill:
This supports providing alternative API servers which follow a slightly different URL scheme, for instance testing or dev API servers which use `/api/development` instead of `/api`.

### DIFF
--- a/seerpy/auth.py
+++ b/seerpy/auth.py
@@ -9,9 +9,10 @@ import requests
 
 class SeerAuth:
 
-    def __init__(self, api_url, email=None, password=None):
+    def __init__(self, api_url, email=None, password=None, dev=False):
         self.api_url = api_url
         self.cookie = None
+        self.dev = dev
 
         self.read_cookie()
         if self.verify_login() == 200:
@@ -48,7 +49,16 @@ class SeerAuth:
         print("login status_code", response.status_code)
         if (response.status_code == requests.codes.ok  # pylint: disable=maybe-no-member
                 and response.cookies):
-            self.cookie = {'seer.sid' : response.cookies['seer.sid']}
+
+            seer_sid = response.cookies.get('seer.sid', False)
+            seerdev_sid = response.cookies.get('seerdev.sid', False)
+
+            self.cookie = {}
+            if seer_sid:
+                self.cookie['seer.sid'] = seer_sid
+            if seerdev_sid:
+                self.cookie['seerdev.sid'] = seerdev_sid
+
         else:
             self.cookie = None
 
@@ -59,7 +69,8 @@ class SeerAuth:
         verify_url = self.api_url + '/auth/verify'
         response = requests.get(url=verify_url, cookies=self.cookie)
         if response.status_code != requests.codes.ok:  # pylint: disable=maybe-no-member
-            print("api verify call returned", response.status_code, "status code")
+            print("api verify call returned",
+                  response.status_code, "status code")
             return 401
 
         json_response = response.json()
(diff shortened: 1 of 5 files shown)


seerpy/auth.py
```diff
@@ -82,10 +93,13 @@ def login_details(self):
             self.email = input('Email Address: ')
             self.password = getpass.getpass('Password: ')
 
+    def get_cookie_path(self):
```
norrishd: I'm definitely glad to have this factored out to a mini helper function rather than hard-coded so many times!

seerpy/auth.py
```diff
@@ -42,24 +43,34 @@ def __init__(self, api_url, email=None, password=None):
...
+        login_url = self.api_url + '/auth/login'
         body = {'email': self.email, 'password': self.password}
         response = requests.post(url=login_url, data=body)
         print("login status_code", response.status_code)
         if (response.status_code == requests.codes.ok  # pylint: disable=maybe-no-member
                 and response.cookies):
-            self.cookie = {'seer.sid' : response.cookies['seer.sid']}
+
+            seer_sid = response.cookies.get('seer.sid', False)
+            seerdev_sid = response.cookies.get('seerdev.sid', False)
```
norrishd: I don't actually know much about the API :-p (e.g. what does 'sid' stand for?). Is the `seerdev.sid` key something you've set up for your dev endpoint, or is it kind of a stable standard that's used?

If the latter, this seems like a pretty good approach. If it's the former, my main thought is "could we generalise this a little more?". E.g. maybe instead of a boolean `dev` flag, the constructor could accept a `cookie_key` string, with a default value of `'seer.sid'`, and this gets used

seerpy/auth.py
```diff
@@ -42,24 +43,34 @@ def __init__(self, api_url, email=None, password=None):
...
+
+            self.cookie = {}
+            if seer_sid:
+                self.cookie['seer.sid'] = seer_sid
+            if seerdev_sid:
+                self.cookie['seerdev.sid'] = seerdev_sid
```
norrishd: It looks from this like the endpoint can return both a `seer.sid` and `seerdev.sid`? If not, this could be `if/elif`, and a very minor thing to note would be that because the _~/seerpy/cookie_ file gets overwritten with the current `self.cookie` whenever there's a successful verification, if you were switching back and forth between dev and production endpoints, you'd have to keep retrieving new cookies and overwriting the other one. If that was an issue, maybe separate files could be saved for the dev vs normal cookies.

seerwill: Good pickup, I'll change to `elif`. 

Regarding your second point, I think as `write_cookie()` and `read_cookie()` use `get_cookie_path()` then separate prod/dev credentials are stored which should mean that both logins can be restored without needing reauth.

seerpy/auth.py
seerwill: I'm fairly certain `sid` stands for `session id` and its part of the API session management code. 

As far as more generalisation, good point. The API expects either of `seer.sid` or `seerdev.sid` cookies depending on the environment the API is running in. No other cookie keys are currently supported. 

Based on this we'd either have to specify `is_dev` as the PR currently does or infer it from the endpoint or from the passed in `cookie_key`. I'd probably argue against passing in a `cookie_key` as I think the user should be more concerned with wether they are running the API against a dev or prod endpoint. It might be a good idea to pull out the current cookie keys into constants so that they can be changed in one place - I'll do that.

It may be useful at a later date to be able to "name" connections (i.e. "dev1" / "prod1") and then store credentials separately for each of those. I'm not sure there is a use case for this at the moment.

norrishd: 100% agree! Nice.

seerpy/auth.py
norrishd: Ah perfect, I missed the two different save paths X-D All good!

PR #47, opened 2020-02-18
Queries and mutations for UserCohorts

seerwill:


seerpy/graphql.py
```diff
@@ -633,3 +633,78 @@ def get_mood_survey_results_query_string(survey_template_ids, limit, offset):
...
+                name
+                users(limit: %0.f, offset: %0.f) {
+                    id
+                }
+            }
+        }
+    """ % (utils.quote_str(user_cohort_id), limit, offset)
+
+
+def create_user_cohort_mutation_string(name, description=None, key=None, user_ids=None):
```
bjdoyle: this should be get_create_user_cohort_mutation_string to be consistent - the first verb is saying we are getting, and everything after that is what we are getting - the create_user_cohort_mutation_string.

seerpy/graphql.py
```diff
@@ -633,3 +633,78 @@ def get_mood_survey_results_query_string(survey_template_ids, limit, offset):
...
+            }) {
+                userCohort {
+                    id
+                }
+            }
+        }
+    """ % (', '.join([f'{key}: {val}' for key, val in args]))
+
+
+def add_users_to_user_cohort_mutation_string(user_cohort_id, user_ids):
```
bjdoyle: get_add_users_to_user_cohort_mutation_string

seerpy/graphql.py
```diff
@@ -633,3 +633,78 @@ def get_mood_survey_results_query_string(survey_template_ids, limit, offset):
...
+                }
+            }
+        }
+    """ % (
+        utils.quote_str(user_cohort_id),
+        get_json_list(user_ids)
+    )
+
+
+def remove_users_from_user_cohort_mutation_string(user_cohort_id, user_ids):
```
bjdoyle: get_remove_users_from_user_cohort_mutation_string

tests/test_seerpy.py
```diff
@@ -537,3 +537,77 @@ def test_generating_remove_studies_from_cohort_mutation(
             }
         }
     """
+
+
+@mock.patch('seerpy.seerpy.GQLClient', autospec=True)
+@mock.patch('seerpy.seerpy.SeerAuth', autospec=True)
+class TestUserCohorts:
+    def test_get_user_ids_in_user_cohort(self, seer_auth, gql_client):
```
bjdoyle: it would be good to add tests where there is no cohort and no users in a cohort.

seerwill: Good point. 

With the current implementation this will throw, as it does with e.g.`client.get_channel_groups('asdf')`. This raises two points:

1) is this how we want to handle `NOT_FOUND` errors? 
2) The throw occurs in the execute_query method which is mocked in the tests. I'm not sure we can mock this throwing an exception? This makes testing quite hard.

Any thoughts?

seerwill: OK, I've updated this commit so that it just handles the NOT_FOUND exception. I'm not sure that we shouldn't be throwing here though as we do elsewhere.

seerpy/seerpy.py
```diff
@@ -95,7 +95,7 @@ def execute_query(self, query_string, party_id=None, invocations=0):
                 self.login()
                 return self.execute_query(query_string, party_id, invocations=invocations)
 
-            raise
+            raise ex
```
bjdoyle: pretty sure this is redundant - a bare raise will rethrow the last exception. I think adding the ex might actually change the stack trace of the exception, so better to leave it off.

seerpy/seerpy.py
```diff
@@ -1023,7 +1023,12 @@ def get_user_ids_in_user_cohort(self, user_cohort_id, limit=200, offset=0):
         while True:
             query_string = graphql.get_user_ids_in_user_cohort_query_string(
                 user_cohort_id, limit, current_offset)
-            response = self.execute_query(query_string)['userCohort']['users']
+
+            try:
+                response = self.execute_query(query_string)['userCohort']['users']
+            except Exception as ex: # pylint: disable=broad-except
+                if 'NOT_FOUND' in str(ex):
+                    break
```
bjdoyle: I think it's better to just let it through - a not found exception is quite acceptable as that is in fact the situation. Catching like this will mean a cohort with no users and a non-existent cohort will have the same return.

tests/test_seerpy.py
bjdoyle: Mocking an exception is pretty easy - just use side_effect - as I think you found?

seerpy/seerpy.py
seerwill: Happy to take it out, however if we aren't doing any special handling of the exception then I don't think we need a test for it?

seerpy/seerpy.py
seerwill: Oops, too much typescript.

tests/test_seerpy.py
```diff
@@ -561,38 +561,6 @@ def test_get_user_ids_in_user_cohort(self, seer_auth, gql_client):
         result = SeerConnect().get_user_ids_in_user_cohort('cohort1')
         assert result == expected_result
```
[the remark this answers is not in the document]
seerwill: As discussed, I have implemented a test to check the NOT_FOUND is passed through to the caller.

PR #48, opened 2020-02-19
Avoid duplicating and / or skipping diary labels

dkeden:
Fix two bugs in fetching diary labels
- BUG: label list was initialised twice if there was more than one label group, FIX: break out of the labelGroup loop after the first initialisation, and check if another query loop is needed (i.e. there are more labels to fetch).
- BUG: if there was more than one label group, labels were skipped for some groups, FIX: move the `offset = offset + limit` outside the labelGroup loop


pkaroly:
Looks good to me! I added a description in case we need to revisit this down the track

PR #51, opened 2020-02-24
Update FitBit data type

pkaroly:
Fix for a bug where Fitbit data timestamps were duplicated because of rounding errors introduced by large floats (posix timestamps in milliseconds)

### Code changes:

- Convert Fitbit data timestamps to python DateTimes to work with timing and duration instead of float arithmetic. Returning dates as the datatype is likely to be less ambiguous for future researchers who use seer-py
- Update the data fields that get queried on a diary channel group by adding a `startTime` so that users know the absolute start of the Fitbit data, and removing some redundant information.

### DIFF
--- a/seerpy/graphql.py
+++ b/seerpy/graphql.py
@@ -519,9 +519,7 @@ def get_diary_study_channel_groups_query_string(patient_id, from_time, to_time):
                     channelGroups {
                         id
                         name
-                        recordsPerChunk
-                        sampleEncoding
-                        compression
+                        startTime
                         segments (ranges: [{ from: %.0f, to: %.0f }]) {
                             id
                             startTime
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -9,6 +9,7 @@ from gql.transport.requests import RequestsHTTPTransport
 import pandas as pd
 from pandas.io.json import json_normalize
 import requests
+from datetime import datetime
 
 from .auth import SeerAuth, COOKIE_KEY_DEV, COOKIE_KEY_PROD
 from . import utils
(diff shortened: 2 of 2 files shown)


seerpy/graphql.py
```diff
@@ -519,9 +519,7 @@ def get_diary_study_channel_groups_query_string(patient_id, from_time, to_time):
                     channelGroups {
                         id
                         name
-                        recordsPerChunk
-                        sampleEncoding
-                        compression
+                        startTime
```
bjdoyle: is this used?

seerpy/seerpy.py
```diff
@@ -840,8 +841,10 @@ def get_diary_fitbit_data(self, segments):
 
         data_list = []
         for idx, url in enumerate(segment_urls):
+            start_time = datetime.utcfromtimestamp(start_times[idx]/1000)
             new_data = utils.get_diary_fitbit_data(url)
-            new_data['timestamp'] = new_data['timestamp'] + start_times[idx]
+            # convert timestamps to true utc datetime
+            new_data['timestamp'] = start_time + pd.to_timedelta(new_data['timestamp'], unit='ms')
```
bjdoyle: so the timestamp is relative to the segment start time?

seerpy/seerpy.py
```diff
@@ -840,8 +841,10 @@ def get_diary_fitbit_data(self, segments):
 
         data_list = []
         for idx, url in enumerate(segment_urls):
+            start_time = datetime.utcfromtimestamp(start_times[idx]/1000)
```
bjdoyle: I notice there is a timezone attribute in the query. Is this used, or is it always 0?
And do you want these times always in UTC regardless of timezone?

seerpy/seerpy.py
pkaroly: correct

seerpy/graphql.py
pkaroly: yes it is the start time of any Fitbit data

seerpy/seerpy.py
pkaroly: Yes the timezone attribute is used to store the device local offset, but the timestamps are in UTC. I want to return them to users in UTC to avoid weirdness with local daylight saving times etc, but the timezone is available if local time is of interest. Most often researchers are more interested in relative times, i.e. event rates and inter-event intervals than true local times, but it's good to have all the info

pkaroly: The general practice we've tried to have for the database is, if there is a timestamp it is in UTC, but it should always be provided with a timezone or offset so that local context is available

PR #52, opened 2020-03-10
Better discoverability re. saving credentials

norrishd:
There is currently no easy way for users to find out about saving their email/password to file to expedite the authentication process.

1. Add a section to the README explaining how authentication works and how you can save a credentials file
2. Print a hint message the first time email/password is manually entered
3. When reading the credentials file, right-strip lines instead of skipping the last character. This is robust to arbitrary whitespace, and means a newline isn't necessary following the password


seerwill:
Is this confusing for our Windows users? i.e. does `~/` work?

norrishd:
> Is this confusing for our Windows users? i.e. does `~/` work?

@seerwill Ooh, potentially confusing! Since the path to the exact file is defined in the function, I've replaced with that. Is it ok with the latest commit?

PR #55, opened 2020-03-18
Add function to request medication compliance from API

dkeden:
- adds function to return medicationCompliance request
- adds function to return medicationCompliance request in DataFrame 

n.b. I've add a `patient_id` at the end of the _dataframe function, similar to `get_diary_labels_dataframe`, although adds to the last column. Can return a sorted or index this column if preferred?

Also, I'm not sure if there's a convention with naming variables so used `medication_compliance` to return the desired response / results, but happy to change

### DIFF
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -583,6 +583,22 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
         label_groups['id'] = patient_id
         return label_groups
 
+    def get_diary_medication_compliance(self, patient_id, from_time=0, to_time=0):
+
+        query_string = graphql.get_diary_medication_compliance_query_string(patient_id, from_time, to_time)
+        response = self.execute_query(query_string)
+
+        return response
+
+    def get_diary_medication_compliance_dataframe(self, patient_id, from_time=0, to_time=0):
+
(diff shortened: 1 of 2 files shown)


seerpy/seerpy.py
```diff
@@ -583,6 +583,21 @@ def get_diary_labels_dataframe(self, patient_id):
         label_groups['id'] = patient_id
         return label_groups
 
+    def get_diary_medication_compliance(self, patient_id, from_time=0, to_time=0):
```
shannonclarke: do you want the default 'to_time' to be 0? normally we use a very large number, to get everything by default, but maybe thats not appropriate for this query?

dkeden: Yes please! I tried with 0 and 9e12, but returns a label for every day since 1970 😄 I think putting the from_time and to_time to 0 returns 'false' in the API and therefore returns the default previous last two weeks

shannonclarke: nice, I figured it was on purpose, but thought I'd check, since its the first one I've seen that's different

bjdoyle: Dominique, maybe you could put the way the times work in a comment so it's clear later on when someone is looking at it?

PR #57, opened 2020-03-24
Change timestamps in fitbit data

rachelstirling:
- Get timezone from segment data
- Make naive timestamps aware
- Change timestamp to local timezone

### DIFF
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -857,9 +857,12 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
         data_list = []
         for idx, url in enumerate(segment_urls):
             start_time = datetime.utcfromtimestamp(start_times[idx]/1000)
+            timezone = segments['segments.timezone'][idx]
             new_data = utils.get_diary_fitbit_data(url)
             # convert timestamps to true utc datetime
             new_data['timestamp'] = start_time + pd.to_timedelta(new_data['timestamp'], unit='ms')
+            new_data['timestamp'] = new_data['timestamp'].dt.tz_localize('UTC')
+            new_data['timestamp'] = new_data['timestamp'].dt.tz_convert('Etc/GMT-{}'.format(timezone))
             new_data['name'] = group_names[idx]
             data_list.append(new_data)
 
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -853,20 +853,17 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
         segment_urls = segments['dataChunks.url']
         group_names = segments['name']
         start_times = segments['segments.startTime']
+        timezones = segments['segments.timezone']
 
         data_list = []
         for idx, url in enumerate(segment_urls):
-            # timestamps are returned in their local times but without their respective timezones
-            start_time = datetime.fromtimestamp(start_times[idx]/1000)
-            # get timezone and - or + sign for local conversion
-            timezone = segments['segments.timezone'][idx]
-            posneg = ('+','-')[timezone > 0]
+            # timestamps are returned in their utc time
+            start_time = datetime.utcfromtimestamp(start_times[idx]/1000)
             new_data = utils.get_diary_fitbit_data(url)
             # convert timestamps to true utc datetime
             new_data['timestamp'] = start_time + pd.to_timedelta(new_data['timestamp'], unit='ms')
-            # add local timezone (this does not change the time, just adds timezone)
-            new_data['timestamp'] = new_data['timestamp'].dt.tz_localize('Etc/GMT{}{}'.format(posneg, timezone))
             new_data['name'] = group_names[idx]
+            new_data['timezone'] = timezones[idx]
             data_list.append(new_data)
 
         if data_list:
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -856,13 +856,16 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
 
         data_list = []
         for idx, url in enumerate(segment_urls):
-            start_time = datetime.utcfromtimestamp(start_times[idx]/1000)
+            # timestamps are returned in their local times but without their respective timezones
+            start_time = datetime.fromtimestamp(start_times[idx]/1000)
+            # get timezone and - or + sign for local conversion
             timezone = segments['segments.timezone'][idx]
+            posneg = ('+','-')[timezone > 0]
             new_data = utils.get_diary_fitbit_data(url)
             # convert timestamps to true utc datetime
             new_data['timestamp'] = start_time + pd.to_timedelta(new_data['timestamp'], unit='ms')
-            new_data['timestamp'] = new_data['timestamp'].dt.tz_localize('UTC')
-            new_data['timestamp'] = new_data['timestamp'].dt.tz_convert('Etc/GMT-{}'.format(timezone))
+            # add local timezone (this does not change the time, just adds timezone)
+            new_data['timestamp'] = new_data['timestamp'].dt.tz_localize('Etc/GMT{}{}'.format(posneg, timezone))
             new_data['name'] = group_names[idx]
             data_list.append(new_data)
 


pkaroly:
Hey @norrishd small PR but I requested your review since Rachel is my PhD student and another set of eyes never hurts..

pkaroly:
> It's difficult to figure out exactly what the `segments` DataFrame returned by `get_diary_channel_groups_dataframe` would contain, and therefore whether that data is permuted correctly in this function. If a few tests could be added for this I'd be _much_ more comfortable. Failing that, could somebody send me an ID for a patient who has some diary data, so I can try this out manually?

I am the best patient: 2b3bcecd-c6a9-471f-a8a6-e3c1e12c6cc8

pkaroly:
> I'm having a bit of trouble doing due diligence in reviewing this! It seems like a lot of diary and Fitbit-related methods have been added to Seerpy, but testing for these new functionalities hasn't been keeping pace.

Completely agreed and so on board! I didn't even realise we had a testing framework in seer-py. I will chat to @dkeden about this and talk about pairing to write some tests for the diary and fitbit functionality. Would you also be a good person to ask about adding new tests? We could organise a Zoom to discuss further..

pkaroly:
> One uncertainty I have is: does this approach work for timezones west of Greenwich? E.g. see - https://stackoverflow.com/questions/4008960/pytz-and-etc-gmt-5.
> 
> I could also be wrong, but is there also a risk that there's a "double counting" thing going on here? I.e. the timezone offset gets added to the start time so that the timestamp is "localized" away from UTC, but then the timezone is specified as UTC. I think that means the `tz_convert` will re-add the offset a second time?

@rachelstirling could you look into this?

norrishd:
@pkaroly Yep I'm all about testing so happy to be involved. I'll check it out again with that ID string, thanks

rachelstirling:
Hi all,
I will try to explain this but it may get a bit confusing. Note here that I am using the original `get_diary_fitbit_data`, not my suggested changes.
When I input start and end times into `get_diary_fitbit_data`, I am entering them as UTC timestamps in milliseconds. The `segments` DataFrame returns segments that contain any data between the start and end time. These come from the Amazon cloud and are stored as UTC timestamps in milliseconds, alongside their respective timezones.
segments
I have attached an example code below:
Example code
In the example, I have set the start time to be Friday March 20 12:00:00 AM and the end time to be Friday March 27 12:00:00 AM. These timestamps are automatically assumed to be UTC, but I have explicitly included the timezone info as UTC for rigour.
The output would be (figure 1):
Figure 1
The data returned appears to be exactly 11 hours behind what I requested, which is the UTC time when Melbourne time is Friday March 20 12:00:00 AM. This is a result of `get_diary_fitbit_data`  where `start_time = datetime.utcfromtimestamp(start_times[idx]/1000)`.
Anyway, it was being converted again to UTC (I am not 100% clear on how this was happening). I can then add the timezone to make each timestamp time aware by using `HRdata['timestamp'] = HRdata['timestamp'].dt.tz_localize('UTC')`. This does not change the times, just adds a timezone. Result shown in Figure 2:
Figure 2
I now need to convert them to their local timezones, which is possible if I pull the timezones from the segments, but the problem is that people can change their timezones between segments so I can't apply the first segments timezone across the whole dataset.
Anyway, what I am trying to say that it is easier to change the timezones if it is done under `get_diary_fitbit_data`, as each segment is being added to the dataset.
With the changes I suggested, the resulting output is:
Figure 3
I have 'confirmed' these times are correct, because HR data and activity data is much lower overnight, whereas previously their heart rates and activity levels were high at night.

In response to your comment about timezones west of Greenwich, I agree that this code does not work in those instances. I have adjusted it and will upload those changes soon.

I hope this all makes sense. If not, we should probably have a meeting on Zoom.

Thanks for all your comments etc.
Rachel

rachelstirling:
I am going to submit another pull request with the following code:

```
    def get_diary_fitbit_data(self, segments):
        """Get fitbit data from a patient's diary study

        Parameters
        ----------
        segments: pandas DataFrame as returned by get_diary_channel_groups_dataframe

        Returns
        -------
        data: pandas DataFrame containing timestamp (adjusted), value, and group name

        """
        segment_urls = segments['dataChunks.url']
        group_names = segments['name']
        start_times = segments['segments.startTime']

        data_list = []
        for idx, url in enumerate(segment_urls):
            # timestamps are returned in their local times but without their respective timezones
            start_time = datetime.fromtimestamp(start_times[idx]/1000)
            # get timezone and - or + sign for local conversion
            timezone = segments['segments.timezone'][idx]
            posneg = ('+','-')[timezone > 0]
            new_data = utils.get_diary_fitbit_data(url)
            # convert timestamps to true utc datetime
            new_data['timestamp'] = start_time + pd.to_timedelta(new_data['timestamp'], unit='ms')
            # add local timezone (this does not change the time, just adds timezone)
            new_data['timestamp'] = new_data['timestamp'].dt.tz_localize('Etc/GMT{}{}'.format(posneg, timezone))
            new_data['name'] = group_names[idx]
            data_list.append(new_data)

        if data_list:
            data = pd.concat(data_list)
        else:
            data = None

        return data
```

This should work for all timezones and does not do any converting between timezones, only adds a timezone. The output is the same when using this code:

Screen Shot 2020-03-27 at 11 58 02 am

norrishd:
Hey @rachelstirling that's the best explanation I've ever seen on a PR, thanks so much! Feel free to open a new PR if you like, otherwise you can just push the changes to this one. Let me know if/when you're done and I'll do a final check and approval

norrishd:
@rachelstirling, it's possible I got confused reading your code, or you may have accounted for it downstream, but in the first part of [this comment](https://github.com/seermedical/seer-py/pull/57#issuecomment-604757047), I'm worried you may have introduced an unintentional time offset where you define `end_time` and set its timezone to be `'UTC'`, based on an unintuitive thin the `datetime` module does under the hood.

To explain what I mean by way of example:

```python
from datetime import datetime
from dateutil import tz  # *See note below

today = datetime.now().date()
today  # == datetime.date(2020, 3, 30)

now_naive = datetime(today.year, today.month, today.day)
now_as_utc = datetime(today.year, today.month, today.day, tzinfo=tz.UTC)
now_aest = datetime(today.year, today.month, today.day, tzinfo=tz.gettz('Australia/Melbourne'))

now_naive.timestamp() - now_aest.timestamp()  # == 0.0
now_naive.timestamp() - now_as_utc.timestamp()  # == -39600.0
datetime.fromtimestamp(now_naive.timestamp())  # == datetime.datetime(2020, 3, 30, 0, 0)
datetime.fromtimestamp(now_aest.timestamp())  # == datetime.datetime(2020, 3, 30, 0, 0)
datetime.fromtimestamp(now_as_utc.timestamp()  # == datetime.datetime(2020, 3, 30, 11, 0)
```

When you call `timestamp()` on a `datetime` object, if it has a timezone defined, `datetime` adds/subtracts the offset to get to UTC, then converts to timestamp (i.e. epoch time, seconds since Jan 1 1970 UTC). When you call `timestamp()` on a datetime object that doesn't have a timezone associated, `datetime` *secretly looks up your system timezone*, applies that to the object, THEN does the conversion to UTC and then to timestamp.

I'm pretty sure that means that when you do `datetime.now()` and set its timezone to UTC, you're actually in effect adding 11 hours to the timestamp. Does that make any sense? Am I totally confused?

Some caveats:
- I know the secret `datetime` timezone conversation happens on Mac OSX Python 3.7, and I'm fairly sure I've tested the same behaviour on Ubuntu, though Windows or different Python versions might handle it differently.
- I'm not sure what your system time is set to. If you're using an EC2 instance, they typically come set to UTC, in which case it shouldn't break anything to specify UTC as you demonstrated in your example.

\* I prefer `dateutil` over `pytz` for the reasons explained in [this post](https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html), but you get the same results with `pytz`.

rachelstirling:
Thanks so much for your thorough analysis. What you are saying seems correct to me, so I think it is best to keep everything in UTC time, as it gets very confusing otherwise. It seems that errors can be introduced quite easily, particularly since `datetime` secretly looks up your system timezone, as you mentioned.
I have changed the PR again to add timezones to the output dataframe (so that I can convert it back to their local times later). Other than that, there are no differences. Let me know if you see any errors with this.
Thank you.

PR #59, opened 2020-04-08
Add function to get createdAt field for diary labels  

dkeden:
This PR adds a function to request the `createdAt` field (from the API) for a given `patient_id`. The `createdAt` field indicates when a diary was created (i.e. when the user signed up).

I have also added the `createdAt` field in the `get_diary_labels` query string, and added a line for the dataframe. It may be redundant to add in both places, but figured that in cases when just the `createdAt` is required, it would be unnecessary to request all diary labels.

New:
seerpy: `get_diary_created_at()`
graphql: `get_diary_created_at_query_string()`

Changes:
seerpy: `get_diary_labels_dataframe()` --> added line
graphql: `get_diary_labels_query_string()` --> added line

### DIFF
--- a/seerpy/graphql.py
+++ b/seerpy/graphql.py
@@ -345,7 +345,7 @@ def get_diary_created_at_query_string(patient_id):
             }
         }""" % patient_id
 
-def get_diary_labels_query_string(patient_id, limit, offset):
+def get_diary_labels_query_string(patient_id, label_type, limit, offset, from_time, to_time, from_duration, to_duration):
     return """
         query {
             patient (id: "%s") {
@@ -359,7 +359,7 @@ def get_diary_labels_query_string(patient_id, limit, offset):
                         labelSourceType
                         name
                         numberOfLabels
-                        labels(limit: %.0f, offset: %.0f) {
+                        labels(limit: %.0f, offset: %.0f, ranges: [{ from: %.0f to: %.0f }, { from: %.0f to: %.0f }]) {
                             id
                             startTime
                             timezone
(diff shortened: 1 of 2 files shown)


seerpy/graphql.py
```diff
@@ -343,7 +352,8 @@ def get_diary_labels_query_string(patient_id, label_type, limit, offset, from_ti
                 id
                 diary {
                     id
-                    labelGroups (filters: [{name: "labelType", value:"%s"}]) {
+                    createdAt
+                    labelGroups {
```
pkaroly, line 356: no filter?

dkeden: Oh man, I must've accepted the wrong change due to conflict. Thanks for picking up on this

dkeden: Hey, sorry. Hopefully that should be fixed up

bjdoyle: I don't see it?

dkeden: Now?

dkeden: i've had some really weird issues w/ github today

bjdoyle: yep all good now. yeah me too, it seems to be struggling. jumping back and forth between multiple PRs also gets me confused!

dkeden: I agree. It was difficult over the break to split these PRs up, but require all of them for dev

dkeden: I'll try putting the next lot into one PR - a bunch of small changes anyway

PR #64, opened 2020-05-01
Add get_patient()

dkeden:
Add function to query patient w/ a given `patient_id`. This function and query is similar to `get_patients()`.

Useful for retrieving the patient's user ID, name and email address.

Being that this is returning user information however, it may be more informative to rename to `get_user()` or `get_user_from_patient_id()`. Thoughts?

### DIFF
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -500,7 +500,7 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
             return orgs
         return pd.DataFrame(orgs)
 
-    def get_user_from_patient_id(self, patient_id):
+    def get_user_from_patient(self, patient_id):
         """
         Get user ID and info from patient ID.
         Parameters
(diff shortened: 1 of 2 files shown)


seerpy/seerpy.py
```diff
@@ -500,6 +500,22 @@ def get_organisations_dataframe(self):
             return orgs
         return pd.DataFrame(orgs)
 
+    def get_patient(self, patient_id):
+        """
+        Get user ID and info from patient ID.
+        Parameters
+        ----------
+        patient_id : str
+            The Seer patient ID for which to retrieve diary labels
```
bjdoyle, line 509: remove Seer, it's implied and a bit confusing.

dkeden: Okay so `get_patient()` is too ambiguous? I'll go with `get_user_for_patient()`. Good idea for the DataFrame, I'll do that now

dkeden: Do you think `get_user_from_patient_id()` would be OK? I think I would remember it better, but that's not a terribly good reason to up vote a suggestion 😄

bjdoyle: I like get_user_for_patient better as it describes what is happening, and the thing you use to get it is the id, but you could potentially do that by some other attribute later if that made sense.

dkeden: Okay, fair enough!

dkeden: Done!

dkeden:
Good to learn some German on the way too. Sounds good. Let me know the way forward for that & if / when I can help

PR #65, opened 2020-05-03
Refactor auth into separate, reusable classes

seerwill:
This provides greater flexibility in logging in using different approaches. Currently only "Base" (no auth) "Default" and "Dev" modes are supported. Existing behaviour is maintained by defaulting to `SeerAuth` where no auth strategy is provided.

- [x] fix mock.patches on tests
- [x] fix / add any required tests
- [x] document

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,8 @@ at https://app.seermedical.com/. You can then use seer-py to authenticate with
 the API.
 
 seer-py will prompt for the username and password that you use to log in to the
-The simplest way to authenticate is to create an instance of `SeerConnect`.
-Seer App.
+Seer App. The simplest way to authenticate is to create an instance of
+`SeerConnect`:
 
 ```
 from seerpy import SeerConnect
--- a/README.md
+++ b/README.md
@@ -32,32 +32,52 @@ To start a Jupyter notebook, run `jupyter notebook` in a command/bash window. Fu
 
 ## Authenticating
 
-To access data from the Seer platform you must first register for an account 
+To access data from the Seer platform you must first register for an account
 at https://app.seermedical.com/. You can then use seer-py to authenticate with
 the API.
 
-Create an instance of the `SeerAuth` class (defined in _auth.py_) and provide
-your email address and password using one of the following options:
+seer-py will prompt for the username and password that you use to log in to the
+The simplest way to authenticate is to create an instance of `SeerConnect`.
+Seer App.
 
-1. Pass them as arguments when instantiating the `SeerAuth` object.
-2. In your home directory, create a  _.seerpy/_ folder, and within it a file
-  named _credentials_. Save your email address on the first line and your
-  password on the second line with no other text.
-3. If you don't use one of the previous options, you will be prompted for your
-  email address and password in the terminal or Jupyter notebook before you can
-  retrieve any data.
+```
+from seerpy import SeerConnect
+
+client = SeerConnect()
+```
+
+Alternatively, you can provide your `email` and `password` as arguments
+`SeerConnect`
+
+```
+client = SeerConnect(email='user@website.com', password='....')
+```
+
+You can also store your credentials by creating a _.seerpy/_ folder in your home
+directory. Within the folder, create a file named _credentials_. Save your email
+address on the first line and your password on the second line with no other
+text. You can then call SeerConnect as above:
+
+```
+client = SeerConnect()
+```
 
 ## Running with other API endpoints
 
-To run seer-py against other environments (for instance against a development API server), perform the following steps:
+To run seer-py in other environments (for instance against a development
+API server), then you can use one of the preconfigured authentication methods:
 
-1. Run the `seer-api` server following the relevant documentation in the API repository
-2. Run a `seer-graphiql` server (which includes the required proxies), ensuring that `HTTPS` is set to `OFF` in the startup script,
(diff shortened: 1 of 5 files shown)


seerpy/auth.py
```diff
@@ -1,26 +1,81 @@
...
 import json
 
 import requests
 
+DEFAULT_COOKIE_KEY = 'seer.sid'
 
-COOKIE_KEY_PROD = 'seer.sid'
-COOKIE_KEY_DEV = 'seerdev.sid'
+
+class BaseConnectionFactory(ABC):
```
bjdoyle, line 13: I don't really like the Factory nomenclature. A Factory tends to refer to a pattern which creates concrete instances of a base class.

seerpy/auth.py
```diff
@@ -1,26 +1,81 @@
...
+
+    @abstractmethod
+    def logout(self):
+        pass
+
+    def get_headers(self):
+        return {}
+
+
+class EmptyConnectionFactory(BaseConnectionFactory):
```
bjdoyle, line 39: I think this should be in the testing code rather than the base module.

seerpy/auth.py
seerwill: Happy either way, in this instance it is referring to the fact that this produces "Connections" that the GQL client uses.

bjdoyle: I feel like you can achieve what you want without using the class hierarchy, and certainly without the abstract base class. I think what you have named DefaultConnectionFactory should just be SeerAuth, and DevConnectionFactory is just a different invocation of that. EmptyConnectionFactory can just be a simple class called something like EmptySeerAuth with the same signature. The only common code is get_connection, which you can share as a mixin. I would rename it to something like get_connection_parameters. An alternative would be to define EmptySeerAuth as the parent and derive SeerAuth from it.

README.md
```diff
@@ -32,32 +32,52 @@ To start a Jupyter notebook, run `jupyter notebook` in a command/bash window. Fu
...
 ## Authenticating
 
-To access data from the Seer platform you must first register for an account 
+To access data from the Seer platform you must first register for an account
 at https://app.seermedical.com/. You can then use seer-py to authenticate with
 the API.
 
-Create an instance of the `SeerAuth` class (defined in _auth.py_) and provide
-your email address and password using one of the following options:
+seer-py will prompt for the username and password that you use to log in to the
```
bjdoyle, line 39: something missing here?

seerwill: Swapped two lines somehow. Fixed.

seerpy/auth.py
```diff
@@ -6,38 +6,112 @@
...
+        pass
+
+    def get_headers(self):
+        return {}
 
 
-class SeerAuth:
+class SeerAuth(BaseAuth):
+    """
+    Creates a default connection factory, which should be used
```
bjdoyle, line 38: change default connection factory to something like "an authorised connection to the seer api"

seerpy/auth.py
```diff
@@ -6,38 +6,112 @@
...
+    for most API use cases.
+    """
 
     help_message_displayed = False
 
-    def __init__(self, api_url, email=None, password=None, dev=False):
-        self.api_url = api_url
+    def __init__(
+            self,
```
bjdoyle, line 45: you're not in typescript world now Will :-)

seerpy/auth.py
```diff
@@ -6,38 +6,112 @@
...
+        home = os.path.expanduser('~')
+        cookie_file = home + self.__get_cookie_path()
+        if os.path.isfile(cookie_file):
+            os.remove(cookie_file)
         self.cookie = None
-        self.dev = dev
 
-        self.read_cookie()
-        if self.verify_login() == 200:
+    def __attempt_login(self):
```
bjdoyle, line 97: change double to single underscore, you're just trying to make them "private" right?

seerpy/auth.py
```diff
@@ -99,34 +153,43 @@ def login_details(self):
             self.email = input('Email Address: ')
             self.password = getpass.getpass('Password: ')
             if not self.help_message_displayed:
-                print(f"\nHint: To skip this in future, save your details to {pswdfile}")
+                print(
+                    f"\nHint: To skip this in future, save your details to {pswdfile}")
```
bjdoyle, line 157: do you have wrapping set to 80? it should be 100.

seerpy/auth.py
```diff
@@ -6,38 +6,112 @@
...
+    def __init__(
+            self,
+            api_url=None,
+            email=None,
+            password=None,
+            cookie_key=DEFAULT_COOKIE_KEY,
+            credential_namespace='cookie'):
+
+        super(SeerAuth, self).__init__(
+            api_url if api_url is not None \
```
bjdoyle, line 53: avoid slash continuance if possible - it looks like this will fit in 100 anyway

seerpy/auth.py
```diff
@@ -6,38 +6,109 @@
...
 
     help_message_displayed = False
 
-    def __init__(self, api_url, email=None, password=None, dev=False):
-        self.api_url = api_url
+    def __init__(self, api_url=None, email=None, password=None, cookie_key=DEFAULT_COOKIE_KEY,
+                 credential_namespace='cookie'):
+
+        super(SeerAuth, self).__init__(
+            api_url if api_url is not None else f"https://api.seermedical.com/api"
```
bjdoyle, line 51: shouldn't be an f string

PR #66, opened 2020-05-03
Return negative of output type instead of None

dkeden:
This PR relates to diary and diaryStudy queries only - modifies output types that are currently None, to the negative instance, e.g. if output is a DataFrame, the negative instance would be an empty DataFrame.

Over the break I also modified the function names related to requested Fitbit data, e.g. from diary_data to diary_study to reflect name in API. I was getting too confused. Think there may have been a discussion re. renaming this though? Can revert these changes if needed


dkeden:
Hey @bjdoyle, I've just merged master into this branch. It was a bit tricky to sort so could use a second pair of eyes. 

I've kept the name changes from 'diary_data' to 'diary_study' to match the API resolver

dkeden:
P.s. I know this is a messy branch arising from the temp. shutdown period

seerpy/seerpy.py
```diff
@@ -1160,9 +1160,11 @@ def get_diary_labels_dataframe(self, patient_id, label_type='all', offset=0, lim
         label_results = self.get_diary_labels(patient_id, label_type, offset, limit, from_time,
                                               to_time, from_duration, to_duration)
         if label_results is None:
-            return label_results
+            return pd.DataFrame({})
```
bjdoyle, line 1163: the empty dict is not necessary, can just be pd.DataFrame().
Same in a few instances below

seerpy/seerpy.py
```diff
@@ -1160,9 +1160,11 @@ def get_diary_labels_dataframe(self, patient_id, label_type='all', offset=0, lim
         label_results = self.get_diary_labels(patient_id, label_type, offset, limit, from_time,
                                               to_time, from_duration, to_duration)
         if label_results is None:
```
bjdoyle, line 1162: I don't think this will ever be true, as get_diary_labels now returns an empty dict instead of None. The check should be if not label_results instead.

seerpy/seerpy.py
```diff
@@ -1160,9 +1160,11 @@ def get_diary_labels_dataframe(self, patient_id, label_type='all', offset=0, lim
...
-            return label_results
+            return pd.DataFrame({})
 
         label_groups = json_normalize(label_results['labelGroups']).sort_index(axis=1)
+        if label_groups.empty:
+            return pd.DataFrame(label_results)
```
bjdoyle, line 1167: I think this was only needed because you slip through the previous check (if label_results is None). I don't think it will ever be true anymore, so is not necessary.

seerpy/seerpy.py
dkeden: Oh! Cool, done

seerpy/seerpy.py
dkeden: Good point, thank you. Done

seerpy/seerpy.py
dkeden: There're some unusual instances where patients have `labelGroups` but no `labels`

dkeden: Done

PR #67, opened 2020-05-04
Quote query string variables inline

bjdoyle:
Moves quoting of string variables into the format string rather than requiring a separate method call.


bjdoyle:
> Is the `utils.quote_string` method used elsewhere or can it be removed altogether?

It's still used in constructing mutation arg lists.

Issue #71, opened 2020-05-07
`get_all_bookings_dataframe()` can break depending on GQL data returned


norrishd:
Steps to reproduce:

```python
from seerpy import SeerConnect
client = SeerConnect()
bookings = client.get_all_bookings_dataframe("9bd42e23-3aee-46e9-81e6-a500aa1c074a",
                                             1588636941143, 1588636941143)
```
Output:
```
TypeError                                 Traceback (most recent call last)
<ipython-input-6-7f83369ec5e2> in <module>
      1 bookings = client.get_all_bookings_dataframe("9bd42e23-3aee-46e9-81e6-a500aa1c074a",
----> 2                                              1588636941143, 1588636941143)

~/miniconda3/envs/watcher/lib/python3.7/site-packages/seerpy/seerpy.py in get_all_bookings_dataframe(self, organisation_id, start_time, end_time)
    693         bookings_response = self.get_all_bookings(organisation_id, start_time, end_time)
    694         bookings = json_normalize(bookings_response).sort_index(axis=1)
--> 695         studies = self.pandas_flatten(bookings, 'patient.', 'studies')
    696         equipment = self.pandas_flatten(bookings, '', 'equipmentItems')
    697         bookings = bookings.drop('patient.studies', errors='ignore', axis='columns')

~/miniconda3/envs/watcher/lib/python3.7/site-packages/seerpy/seerpy.py in pandas_flatten(parent, parent_name, child_name)
    110         for i in range(len(parent)):
    111             parent_id = parent[parent_name+'id'][i]
--> 112             child = json_normalize(parent[parent_name+child_name][i]).sort_index(axis=1)
    113             child.columns = [child_name+'.' + str(col) for col in child.columns]
    114             child[parent_name+'id'] = parent_id

~/miniconda3/envs/watcher/lib/python3.7/site-packages/pandas/io/json/_normalize.py in json_normalize(data, record_path, meta, meta_prefix, record_prefix, errors, sep, max_level)
    256 
    257     if record_path is None:
--> 258         if any([isinstance(x, dict) for x in y.values()] for y in data):
    259             # naive normalization, this is idempotent for flat records
    260             # and potentially will inflate the data considerably for

TypeError: 'float' object is not iterable
```

`get_all_bookings()` still works. I believe the problem may be the `pandas_flatten` expects a `pd.DataFrame` column where all cells have lists of dict values, and it might be hitting an `np.NaN` instead.

PR #74, opened 2020-05-13
Fix TypeError when calling `pandas_flatten()` on DF with missing values

norrishd:
Calling `SeerConnect.pandas_flatten` raises a `TypeError` when the `child_name` column of the `parent` DataFrame has any missing values. E.g. Calling `SeerConnect.get_all_bookings_dataframe()` currently hits this issue, as there are recent bookings with no `patient.studies` value.

The exception is thrown by Pandas trying to iterate over a non-iterable (`math.nan` values in the case of the bookings data). As such, skip cells that don't contains a list.

Not sure if this is the best solution, e.g. should we also allow tuples or iterables generally using `collections.Iterable`?

I also tried:
```python
try:
    child = json_normalize(cell_to_flatten).sort_index(axis=1)
except TypeError:
```
However this doesn't catch the internal Pandas exception.

And:
```python
if math.isnan(cell_to_flatten):
    continue
```
However this raises a `TypeError` when it gets a list (i.e. what's expected) rather than a number/NaN.

### DIFF
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -254,13 +254,15 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
         child_list = []
         for i in range(len(parent)):
             parent_id = parent[parent_name+'id'][i]
-            child = json_normalize(parent[parent_name+child_name][i]).sort_index(axis=1)
-            child.columns = [child_name+'.' + str(col) for col in child.columns]
-            child[parent_name+'id'] = parent_id
-            child_list.append(child)
+            cell_to_flatten = parent[parent_name+child_name][i]
+            if isinstance(cell_to_flatten, list):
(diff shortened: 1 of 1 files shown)


bjdoyle:
So you just end up with empty columns for the entries that are blank? (Assuming there is at least one non-blank cell)

norrishd:
> So you just end up with empty columns for the entries that are blank? (Assuming there is at least one non-blank cell)

Yep. There won't be a flattened DataFrame created for those entries to merge back into the parent one, so Pandas either inserts np.Nan or float.nan (I forget which)

bjdoyle:
So this fixes https://github.com/seermedical/seer-py/issues/71 right?

norrishd:
> So this fixes #71 right?

Oh yeah! Thanks for reminding me, I meant to link that.

PR #78, opened 2020-06-05
Add function to retrieve patient insights / report

dkeden:
Adds function to retrieve patient insights from API. 

May have features added in the future, e.g. filters / ranges / retrieve_last, etc. in the future, but this is an initial implementation.


seerpy/seerpy.py
```diff
@@ -1036,6 +1036,27 @@ def get_documents_for_studies_dataframe(self, study_ids, limit=50):
...
+        offset : int, optional
+            Optional index of first record to return
+        Returns
+        -------
+        result : dict
+            Returns a dictionary patient insights.
+        """
+        query_string = graphql.get_insights_query_string(patient_id, limit, offset)
+        response = self.execute_query(query_string)
+        return response['patient']
```
bjdoyle, line 1058: Take a look at any of the functions which use get_paginated_response, like get_mood_survey_results. If you change your query string a bit and get_paginated_response you'll get pagination.

dkeden: Ah! Will do!

dkeden: I completely overlooked the pagination

dkeden: P.s. I've snuck in a docstring change to `get_diary_medication_alerts()` rather than creating a new branch. Is that alright?

dkeden: DW! I'll put in a separate PR

PR #79, opened 2020-06-09
Remove default options from pylintrc

bjdoyle:
I have removed default options from the pylintrc to make it easier to see what we have set explicitly.
The options that remain are mostly other than the default values set by pylint, except for some (like format and naming conventions) which are left there for reference, and the design section, at least some of which is probably worth revisiting to see if we should change the values or ditch the checks entirely.
I would like us to use mostly the same pylint settings across projects, except where there is a need for specific settings in particular projects. For example, I think we can probably re-instate the docstring warning for seer-py, but I don't think that is appropriate across the board.

If there are any particular settings anyone has issues with, please let me know.
Though I don't think we need to settle on any potential changes to the design settings in this PR.

I have not made any changes to the code to fix any pylint errors, that's for a separate PR.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -29,7 +29,6 @@ disable=print-statement,
         raw-checker-failed,
         bad-inline-option,
         locally-disabled,
-        locally-enabled,
         file-ignored,
         suppressed-message,
         useless-suppression,
@@ -97,14 +96,6 @@ disable=print-statement,
         exception-escape,
         comprehension-escape,  # messages after here are added for this project
         missing-docstring
-
-
-[TYPECHECK]
-
-# List of members which are set dynamically and missed by pylint inference
-# system, and so shouldn't trigger E1101 when accessed. Python regular
-# expressions are accepted.
-generated-members=torch.*
 
 
 [FORMAT]
--- a/pylintrc
+++ b/pylintrc
@@ -1,55 +1,11 @@
 [MASTER]
 
-# A comma-separated list of package or module names from where C extensions may
-# be loaded. Extensions are loading into the active Python interpreter and may
-# run arbitrary code.
-extension-pkg-whitelist=
-
-# Add files or directories to the blacklist. They should be base names, not
-# paths.
-ignore=CVS
-
-# Add files or directories matching the regex patterns to the blacklist. The
-# regex matches against base names, not paths.
-ignore-patterns=
-
-# Python code to execute, usually for sys.path manipulation such as
-# pygtk.require().
-#init-hook=
-
-# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
-# number of processors available to use.
-jobs=1
-
-# Control the amount of potential inferred values when inferring a single
-# object. This can help the performance when dealing with large functions or
-# complex, nested conditions.
-limit-inference-results=100
-
-# List of plugins (as comma separated values of python modules names) to load,
-# usually to register additional checkers.
-load-plugins=
-
-# Pickle collected data for later comparisons.
-persistent=yes
-
-# Specify a configuration file.
-#rcfile=
-
-# When enabled, pylint would attempt to guess common misconfiguration and emit
-# user-friendly hints instead of false-positive error messages.
-suggestion-mode=yes
-
-# Allow loading of arbitrary C extensions. Extensions are imported into the
-# active Python interpreter and may run arbitrary code.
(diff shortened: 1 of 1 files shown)


pylintrc
```diff
@@ -139,185 +94,12 @@ disable=print-statement,
...
-
-
-[REPORTS]
-
-# Python expression which should return a note less than 10 (10 is the highest
-# note). You have access to the variables errors warning, statement which
-# respectively contain the number of errors / warnings messages and the total
-# number of statements analyzed. This is used by the global evaluation report
-# (RP0004).
-evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / statement) * 10)
```
norrishd, line 159: Will this break the code scores that Pylint outputs? I find these mildly useful (or realistically maybe just rewarding to see the score go up when you fix issues)

bjdoyle: That's the default pylint evaluation function.
Pretty much everything that has been removed is just the pylint default.
I removed it so you see what we have explicitly set.

pylintrc
```diff
@@ -139,185 +94,12 @@ disable=print-statement,
...
-# it will be considered as an explicit return statement and no message will be
-# printed.
-never-returning-functions=sys.exit
-
-
-[LOGGING]
-
-# Logging modules to check that the string format arguments are in logging
-# function parameter format.
-logging-modules=logging
```
norrishd, line 193: I think this is a mildly useful thing to call out - 
https://reinout.vanrees.org/weblog/2015/06/05/logging-formatting.html

pylintrc
```diff
@@ -139,185 +94,12 @@ disable=print-statement,
...
-
-# List of decorators that produce context managers, such as
-# contextlib.contextmanager. Add to this list to register other decorators that
-# produce valid context managers.
-contextmanager-decorators=contextlib.contextmanager
-
-# List of members which are set dynamically and missed by pylint inference
-# system, and so shouldn't trigger E1101 when accessed. Python regular
-# expressions are accepted.
-generated-members=
```
norrishd, line 234: I reckon we should actually retain this and use it to get rid of those annoying messages about dynamically generated `boto3` classes not being valid:

https://github.com/boto/botocore/issues/1870

bjdoyle: yeah I use this in algo for torch, but there's nothing in seerpy that uses it.
Not sure if it's better to have it in here or not?

pylintrc
norrishd: (Oh cool, I didn't really grok that all these options would all still work even if not explicitly included in the pylintrc)

pylintrc
norrishd: Ah ok, I didn't really grok that all these options would all still work even if not explicitly included in the pylintrc

pylintrc
norrishd: If this is kind of going to be our default/template pylintrc, I'd lean towards yes. It will just mean one less thing to think about/change when starting a new project

norrishd: But yeah, either argument seems reasonable

pylintrc
```diff
@@ -342,185 +123,51 @@ single-line-if-stmt=no
...
-#function-rgx=
-
 # Good variable names which should always be accepted, separated by a comma.
 good-names=i,
            j,
            k,
            ex,
            Run,
-           _,
+           _,  # items after this are added for this project
```
norrishd, line 156: My last job used `'df'` for a generic pandas dataframe, e.g. as a function argument so I'm kind of used to that, but also get that it can be problematically ambiguous

pylintrc
bjdoyle: It will pick up any usage that doesn't conform to that anyway

pylintrc
bjdoyle: yeah I deliberately left that off. It's okay in a one-off script, but in anything other than that, it should be named after what's in it.

pylintrc
bjdoyle: I was on the fence, but yeah I'll add it back in and put torch and boto in as standard.

bjdoyle: okay I added generated-members back, but boto I think actually needs to be in ignored-modules as it is actually multiple classes within boto which are problematic so I've added that as well. Let me know whether that works in you situation or not @norrishd

norrishd: Yep this fixes the no-member complaint!

PR #80, opened 2020-06-15
add medication alert windows

dkeden:
Adds query for retrieving medication alert windows w/ `isActive` filter. 

No more tripping, works as expected.


seerpy/seerpy.py
```diff
@@ -1217,9 +1219,30 @@ def get_diary_medication_alerts_dataframe(self, patient_id, from_time=0, to_time
         labels = self.pandas_flatten(alerts, '', 'labels')
         return labels
 
+    def get_diary_medication_alert_windows(self, patient_id, is_active="both"):
```
seerwill, line 1222: The valid values for `is_active` are `True` or `False` (or omitted altogether). I haven't checked but I'm guessing the error will be due to filter validation?

The query here should probably default to `None`, and the query_string method will need to omit the filter entirely if the arg is `None`.

dkeden: I thought they were `"true"` `"false"` and "`both"`

dkeden: Happy to set to `None` but would require a few extra lines. Figured that "both" returns the default anyway

dkeden: i.e. is equivalent to no filter

bjdoyle: 1. yes, is_active as the parameter name is fine
2. I'd have to see the valid queries and what this is producing which is invalid, but assuming the above is true then I'd agree with Will's assessment.

bjdoyle: Also, I commented before I saw you're comments, so apologies if it seems like I ignored them.

bjdoyle: Also, I just wrote you're, then edited it to correct it to your, but feel so ashamed that I feel the need to admit to my misdeeds.

dkeden: Oh all good, I didn't notice anything

seerpy/graphql.py
```diff
@@ -505,6 +505,27 @@ def get_diary_medication_alerts_query_string(patient_id, from_time, to_time):
...
+                patient (id: "%s") {
+                    diary {
+                        id
+                        alerts {
+                            id
+                            name
+                          	windows %s {
+                              startTime
+                              timezone
+                              endTime
```
seerwill, line 520: I think indentation is a little messed up here

seerpy/seerpy.py
```diff
@@ -1217,9 +1219,34 @@ def get_diary_medication_alerts_dataframe(self, patient_id, from_time=0, to_time
...
+            with a 'windows' key that indexes list of dict with keys 'startTime',
+            'timezone', and 'endTime'.
+        """
+        if is_active is not None:
+            filter_string = f'(filters: [{{name: "isActive", value: "{"true" if is_active else "false"}"}}])'
+        else:
+            filter_string = ''
+        query_string = graphql.get_diary_medication_alert_windows_query_string(
+            patient_id, filter_string)
+        return self.execute_query(query_string)['patient']['diary']['alerts']
```
seerwill, line 1245: What happens if the query returns no data? Is there an exception raised?

dkeden: I don't think there's ever no data because we all have `default_prn` by default

seerpy/graphql.py
dkeden: Oops, fixed!

seerpy/graphql.py
```diff
@@ -515,9 +515,9 @@ def get_diary_medication_alert_windows_query_string(patient_id, filter_string):
...
                             name
                           	windows %s {
-                              startTime
-                              timezone
-                              endTime
+                                startTime
+                                timezone
+                                endTime
                             }
                         }
```
bjdoyle, line 523: you still have a blank line at 523

dkeden: Removed

PR #82, opened 2020-06-16
fix add label using graphql variables

bjdoyle:
Matias had a problem adding labels with "\n" in them - the way we were (not) encoding them meant that we were submitting non-conformant graphql.
I have changed the add_label mutation to use graphql variables (thanks @seerwill) instead of directly inserting the strings into the query.
This is something we should do elsewhere as well, but I just wanted to get this change in for Matias, and also to make the change manageable.

### DIFF
--- a/seerpy/graphql.py
+++ b/seerpy/graphql.py
@@ -1,9 +1,80 @@
 """
 GraphQL queries used by various seerpy.SeerConnect methods.
 """
-import json
-
 from . import utils
+
+
+def get_json_list(list_of_strings, include_brackets=True):
+    """
+    Convert a list of strings into a single string representation, suitable for
+    inclusion in GraphQL queries.
+
+    Parameters
+    ---------
+    list_of_strings : list of str
+        Strings to convert to single string
+    include_backets : bool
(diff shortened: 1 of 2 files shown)


seerpy/graphql.py
```diff
@@ -281,18 +281,15 @@ def get_studies_by_study_id_paged_query_string(study_ids):
         }}}}"""
 
 
-def get_add_labels_mutation_string(group_id, labels):
-    labels_string = get_string_from_list_of_dicts(labels)
-
+def get_add_labels_mutation_string():
```
seerwill, line 284: Seems a bit strange having this as a function given it just returns a constant string but probably not worth changing in this instance to keep a common interface.

I think if we do a large refactor to use query variables at some point we should consider creating these as constants, and wrapping them in `gql` "tags", e.g. https://github.com/graphql-python/gql/blob/master/gql-checker/tests/test_cases/validation.py

bjdoyle: Yeah just wanted to change the minimum amount for this change. Yeah should add some tests, but I did test the change against production to confirm that it works.

PR #86, opened 2020-06-25
Don't clear password unless get 401 error

norrishd:
If authentication fails, only ask user to re-enter password/read password from file again if the status code returned was a 401 (unauthorized).

For any other "non-ok" status code (e.g. 429 timeout, 502 bad gateway), for each retry sleep for an exponentially increasing period of time then try again with same password.

Sleep time calibrated to be roughly 5/20/60 seconds, with a random few seconds added to avoid the [thundering heard problem](https://www.apollographql.com/docs/link/links/retry/#avoiding-thundering-herd).

### DIFF
--- a/seerpy/auth.py
+++ b/seerpy/auth.py
@@ -6,6 +6,7 @@ Copyright 2017 Seer Medical Pty Ltd, Inc. or its affiliates. All Rights Reserved
 import getpass
 import os
 import json
+import time
 
 import requests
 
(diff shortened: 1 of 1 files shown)


norrishd:
Though I'd make this a draft PR to get feedback on general approach/how long to sleep for etc. before flagging it as ready for review.

NB I changed `_verify_login()` to pass on the status code rather than turning everything into a 401.

seerpy/auth.py
```diff
@@ -121,16 +122,19 @@ def _attempt_login(self):
...
                       'app.seermedical.com to reset your password')
+                raise InterruptedError('Authentication Failed')
+
+            if response == 401:
+                print('\nLogin error, please re-enter your email and password: \n')
                 self.cookie = None
                 self.password = None
-                raise InterruptedError('Authentication Failed')
+            else:
+                sleep_time = (allowed_attempts + 3)**2
```
bjdoyle, line 135: The sleep in seerpy is designed to increase each time if there are multiple failures. This is just 36 each time. I would either work it off the number of attempts (i) so that it increases each time, or just make it explicitly a constant - like 30 or whatever.

norrishd: Oops yeah, this should've been `i` not `allowed_attempts`, with the intention being that it would sleep for 9, 16 then 25 seconds before ultimately failing.

That's a very random choice though, not sure what more implementation-informed values might be. @seerwill do you know if the API side prescribes waiting a certain duration after a failure (especially a 502)?

norrishd:
Confession: ended up using https://mycurvefit.com/ to get the equation for the increasing sleep times that felt logical to me.

PR #90, opened 2020-07-07
Add api key support

bjdoyle:
Adds support for api keys and the new SDK servers.

Will read an api key from ~/.seerpy/seerpy*.pem if available.
Will also optionally get the id and region from the key file name.

Defaults to api key if an api key file is found, otherwise will use email authentication.

There is also a use_email flag which will force use of email if True.


seerpy/seerpy.py
```diff
@@ -57,7 +57,7 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
     graphql_client = None
 
     def __init__(self, api_url='https://api.seermedical.com/api', email=None, password=None,
```
robrkerr, line 59: Is it possible to not have a default value for the `api_url` here because currently this would always be used over the different defaults for each of the authentication classes?

robrkerr: I guess not a big deal as you can just pass in the `auth` object itself.

seerpy/auth.py
```diff
@@ -32,10 +33,10 @@ def get_connection_parameters(self, party_id=None):
             'timeout': 30
         }
 
-    def login(self):
-        pass
+    def handle_query_error_pre_sleep(self, error_string):  # pylint: disable=unused-argument
+        return True
 
-    def logout(self):
+    def handle_query_error_post_sleep(self, error_string):
```
robrkerr, line 39: Does this not need the same `# pylint: disable=unused-argument` comment?

seerpy/seerpy.py
seerwill: From memory this was about maintaining backwards compatibility - i.e. existing scripts can continue to instantiate SeerConnect the same way.

requirements.txt
```diff
@@ -7,4 +7,5 @@ scipy
 # TODO: graphql-core 3.0.0 has a problem as at 02/12/2019
 # revisit in the future, we may be able to remove the restriction
 graphql-core<3.0.0
+pyjwt
```
seerwill, line 10: Does RS256 require the crypto extension? If so from the [pyjwt docs](https://pyjwt.readthedocs.io/en/latest/installation.html):

> The pyjwt[crypto] format is recommended in requirements files in projects using PyJWT, as a separate cryptography requirement line may later be mistaken for an unused requirement and removed.

seerpy/auth.py
```diff
@@ -204,6 +215,42 @@ def _read_cookie(self):
                 self.cookie = json.loads(f.read().strip())
 
 
+class SeerApiKeyAuth(BaseAuth):
+    """
+    Creates an authenticated connection to the Seer API using an API key. This will become the
+    default for most use cases.
+    """
+    def __init__(self, api_key_id, api_key_path, api_url=None):
```
seerwill, line 223: I find it a bit confusing having the order of the arguments reversed here compared to other Auth providers, but given the optional pattern not sure of another way.

seerpy/auth.py
```diff
@@ -204,6 +215,42 @@ def _read_cookie(self):
...
+            The UUID for a Seer api key
+        api_key_path : str
+            The path to a Seer api key file
+        api_url : str, optional
+            Base URL of API endpoint
+        """
+
+        # TODO: not sure what the api url actually is
+        super(SeerApiKeyAuth,
+              self).__init__(api_url if api_url is not None else "https://sdk.seermedical.com/api")
```
seerwill, line 239: Could this be simplified to the following (and in SeerAuth above) by setting `api_url='https://......'` in the function args?

```suggestion
              self).__init__(api_url)
```

seerpy/auth.py
```diff
@@ -204,6 +215,42 @@ def _read_cookie(self):
...
+
+        self.api_key_id = api_key_id
+        # TODO: should we default the path to something like "~/.ssh/seerpy.pem"?
```
seerwill, line 242: Could follow the `_read_cookie()` pattern from `SeerAuth`?

seerpy/seerpy.py
```diff
@@ -136,16 +140,16 @@ def execute_query(self, query_string, party_id=None, invocations=0, variable_val
                 raise
             error_string = str(ex)
             if any(api_error in error_string for api_error in resolvable_api_errors):
-                if 'NOT_AUTHENTICATED' in error_string:
-                    self.seer_auth.logout()
-                else:
+                if not self.seer_auth.handle_query_error_pre_sleep(error_string):
```
seerwill, line 143: I don't know if it's worth handling separately but the "best" handling of errors is likely to be different for API key auth compared to SeerAuth. For instance `NOT_AUTHENTICATED` is probably not recoverable with API key authentication.

One alternative may be to add a `def can_retry_authentication(self, error_string) -> boolean` function to `BaseAuth`, or maybe even simpler to raise in `SeerApiKeyAuth.handle_query_error_pre_sleep` if the error string is `Not Authenticated`.

seerpy/auth.py
bjdoyle: No, pylint seems to realise what's going on from pass being the entire body.

seerpy/auth.py
bjdoyle: Yeah, not ideal, but an artefact of the differing optionality.

bjdoyle: And again, the other will probably go away.

seerpy/seerpy.py
bjdoyle: This was just to move the login stuff into the auth class - I don't think any other implantation needs to do anything here - and I think this can be removed if we no longer support login.

requirements.txt
bjdoyle: possibly, find out when we test it

seerpy/auth.py
bjdoyle: It was more a question of do we want to? Another possibility is to use the api key to select a specific pem file with the key as part of the name if no pem file is specified.

seerpy/seerpy.py
bjdoyle: True. It probably is better to default it to none. The defaults are in the auth classes, but we can leave the parameter so people still have the option to pass an alternative in and use the same auth.

bjdoyle: Actually that won't work, then it passes none in to the child classes, and I've just made the other change Will suggested! Now I've realised why the auth classes treated api_url like that. I'll revert that change and set it to none here. I think.

bjdoyle: I done that ^.

seerpy/auth.py
bjdoyle: As it turns out, no - you did it that way for a reason!

tests/test_seerpy.py
```diff
@@ -26,7 +25,7 @@
 @mock.patch('seerpy.seerpy.SeerAuth', autospec=True)
 class TestSeerConnect:
     def test_success(self, seer_auth):
-        seer_auth.return_value.cookie = {DEFAULT_COOKIE_KEY: "cookie"}
+        seer_auth.return_value.cookie = {'seer.sid': "cookie"}
```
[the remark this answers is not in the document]
bjdoyle: Done.

seerpy/auth.py
seerwill: Storing by ID could support trying with different creds which might be useful?

seerpy/seerpy.py
seerwill: I think the point here is that there isn't much point retrying API key auth if "NOT_AUTHENTICATED" occurs, i.e. I think this should be an unrecoverable error for API keys but should trigger a re-login for user/pass?

bjdoyle: It doesn't do anything for api key - the login auth is the only one that implements this.

bjdoyle: Oh I see what you mean, we should raise in that case. Yeah. I think eventually it would be removed from resolvable_api_errors but in the meantime yeah I'
ll do that.

requirements.txt
seerwill: Seems to work fine without it locally.

seerpy/auth.py
```diff
@@ -47,27 +84,29 @@ class SeerAuth(BaseAuth):
...
-            Flag to query the dev rather than production endpoint
+        cookie_key : str, optional
+            ?
+        credential_namespace : str, optional
+            ?
         """
 
-        super(SeerAuth,
-              self).__init__(api_url if api_url is not None else "https://api.seermedical.com/api")
+        super(SeerAuth, self).__init__(api_url if api_url else 'https://api.seermedical.com/api')
```
bjdoyle, line 109: I'm unsure how this url changes for different regions, but we could use the region parameter to get_parameters to vary it without supplying the whole url.

requirements.txt
bjdoyle: BTW @seerwill I had to add the crypto extension - maybe you already had it from something else?

seerpy/auth.py
bjdoyle: @seerwill - I implemented this, it looks in ~/.seerpy for anything following the pattern seerpy*.pem. You can also store the id, the region, and also a default indicator, e.g. seerpy.default.id.uk.pem would use this as the default (if there is more than one) and set the id to is and the region to uk.

tests/test_auth.py
```diff
@@ -14,6 +16,77 @@
...
+        # check result
+        assert result == expected_return
+
+    @mock.patch.object(SeerAuth, '__init__', autospec=True, return_value=None)
+    def test_use_email(self, seer_auth_init, mock_glob):
+        # setup
+        mock_glob.return_value = ['seerpy.pem']
+
+        # run test
+        result = auth.get_auth(api_key_id='api_key_id', api_key_path='api_key_path', use_email=True)
```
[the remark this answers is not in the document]
bjdoyle: Because you might have email credentials in a file and api keys in files, in which case it will default to the api key file. use_email=True lets you use the credentials, or lets you use the command line for email credentials. Providing the key details is just to test that use_email will override regardless.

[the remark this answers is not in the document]
bjdoyle: A lot of this would simplify if we no longer support email access.

bjdoyle:
@seerwill Yeah, it's complex. But you either do it all or you have to stop somewhere and not support all options. Many of them may be unused by most people. But what I've written will cover all bases (I think), all paths are tested, and I couldn't see any obvious options to disallow that would significantly simplify things.

seerpy/auth.py
```diff
@@ -3,17 +3,54 @@
...
+    # provide nothing - if one file or one contains default, must contain id, can contain region
+    # seerpy.region.id.pem
+    # seerpy.default.region.id.pem
+
+    # provide id - if file matching id found, or single file, can include region
+    # seerpy.id.pem
+    # seerpy.pem
+
+    # provide region - if file matching region found - must contain id
+    # seerpy.region.id.pem
```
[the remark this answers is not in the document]
bjdoyle: converted

tests/test_seerpy.py
```diff
@@ -18,61 +18,57 @@
...
+def seer_connect():
+    return SeerConnect(seer_auth=auth.BaseAuth(api_url=''))
 
-        result = SeerConnect()
+
+class TestSeerConnect:
+    def test_success(self):
+        result = SeerConnect(seer_auth=auth.BaseAuth(api_url=''))
 
         assert result.graphql_client
```
seerwill, line 38: I'm not familiar with pytest fixtures, but is this equivalent to

```suggestion
        assert seer_connect.graphql_client
```

?

bjdoyle: No, that test doesn't use a fixture, the fixture just happens to have the same result as this test.
I used BaseAuth for both because creating one doesn't have any side effects that I don't care about, so I don't have to mock anything to simplify it.
This test is checking the init, that if you pass a valid auth, then everything works, and it checks that by checking that the graphql_client has a a value.
The fixture is used in other tests that aren't testing init, but that want a setup SeerConnect to test other methods on it. If you check most other test methods, they have a final parameter named seer_connect, that doesn't correspond to any passed in mocks. If there is a missing parameter, Mock looks for a fixture of the same name and passes the return value of that function as the parameter.
That replaces a lot of the patches and mock value setting that was there previously.

bjdoyle: BTW if you ever come across the error "No fixture named X" when you've got the wrong number of parameters for a test method compared to the number of patches, now you why you get that message instead of something about patches or mocks.

seerwill: Ah ok, thanks for the explanation!

bjdoyle:
Thanks @seerwill.
I have another small change to make to support passing in an api_key itself rather than a file path, to support use in lambdas. I'll do that as another small PR. Then, yeah, I think a new release number is warranted, good idea.
Getting back to how we read in pem files, I'm very open to seeing how it gets used and potentially tweaking it / simplifying it if we don't need all of the options. I've tried to cater for the use cases I could think of, but what is actually useful we will figure out as we start to use it more. Some of the options are useful for the situation where you have more than one key for a region, and use multiple regions with different keys, and I'm not sure if that is actually going to happen or not, but we'll see.

bjdoyle:
Oh yeah and those test warnings I think come from scikit. We probably need to do a version update.

PR #91, opened 2020-07-14
Add ability to download timestamped Sense data

bjdoyle:
Data from the sense is stored in a different format to that which comes from the Siesta.
This PR adds the "timestamped" flag from the channel group, which indicates whether the data is in this format or the existing format, and changes the utils.download_channel_data function to cater for the new format as well as the existing one.

Again, there is no unit test for this function as yet, but I have tested it on both Sense and Siesta data.

### DIFF
--- a/seerpy/utils.py
+++ b/seerpy/utils.py
@@ -60,11 +60,19 @@ def download_channel_data(data_q, download_function):
         data_type = meta_data['channelGroups.sampleEncoding']
         data = np.frombuffer(data, dtype=np.dtype(data_type))
         data = data.astype(np.float32)
-        data = data.reshape(-1, len(channel_names),
-                            int(meta_data['channelGroups.samplesPerRecord']))
-        data = np.transpose(data, (0, 2, 1))
-        data = data.reshape(-1, data.shape[2])
-        if 'int' in data_type:
+
+        column_names = channel_names
+
+        if meta_data['channelGroups.timestamped']:
+            column_names = ['time'] + channel_names
+            data = data.reshape(-1, len(column_names))
+        else:
+            data = data.reshape(-1, len(channel_names),
+                                int(meta_data['channelGroups.samplesPerRecord']))
+            data = np.transpose(data, (0, 2, 1))
+            data = data.reshape(-1, data.shape[2])
+
+        if 'int' in data_type:  # TODO: why is this only required for ints?
             nan_mask = np.all(data == np.iinfo(np.dtype(data_type)).min, axis=1)
             if nan_mask[-1]:
                 nan_mask_corrected = np.ones(nan_mask.shape, dtype=bool)
@@ -77,11 +85,9 @@ def download_channel_data(data_q, download_function):
 
             # fill missing values with nans
             data[np.all(data == np.iinfo(np.dtype(data_type)).min, axis=1), :] = np.nan
-        # TODO: what happens for floats?
-        chan_min = meta_data['channelGroups.signalMin'].astype(np.float64)
-        chan_max = meta_data['channelGroups.signalMax'].astype(np.float64)
-        exponent = meta_data['channelGroups.exponent'].astype(np.float64)
-        if 'int' in data_type:
+
+            chan_min = meta_data['channelGroups.signalMin'].astype(np.float64)
+            chan_max = meta_data['channelGroups.signalMax'].astype(np.float64)
             chan_diff = chan_max - chan_min
             dig_min = np.iinfo(data_type).min
             dig_max = np.iinfo(data_type).max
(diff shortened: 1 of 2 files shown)


seerpy/utils.py
```diff
@@ -60,11 +60,24 @@ def download_channel_data(data_q, download_function):
...
+            data = data.reshape(-1, len(column_names))
+        else:
+            # EDF data is in the format [record 1: (ch1 sample1, ch1 sample2, ..., ch2 sampleN),
+            # (ch2 sample1, ch2 sample2, ..., ch2 sampleN), ...][record2: ...], ..., [recordN: ...]
+            data = data.reshape(-1, len(channel_names),
+                                int(meta_data['channelGroups.samplesPerRecord']))
+            data = np.transpose(data, (0, 2, 1))
+            data = data.reshape(-1, data.shape[2])
+
+        if 'int' in data_type:  # TODO: why is this only required for ints?
```
norrishd, line 80: That's a very good and perplexing question. Am I understanding correctly that this logic branch drops all rows where all channels have the lowest possible value (I guess this is used as a stand-in for NaN), starting at the last row and working backwards until it hits a row with normal data?

Not sure what other values are possible for `data_type`, or if there's some reason why only 'int' encoded data would have extraneous rows? Shrug.

seerpy/utils.py
```diff
@@ -90,22 +101,32 @@ def download_channel_data(data_q, download_function):
...
         data['channelGroups.id'] = channel_groups_id
         data['segments.id'] = segments_id
         data = data[['time', 'id', 'channelGroups.id', 'segments.id'] + channel_names]
 
         # data chunks seem to always be 10 seconds even if they don't contain that much data
-        segment_start = meta_data['segments.startTime']
-        segment_end = segment_start + meta_data['segments.duration']
-        data = data[(data['time'] >= segment_start) & (data['time'] < segment_end)]
+        segment_end = meta_data['segments.startTime'] + meta_data['segments.duration']
+        data = data[data['time'] < segment_end]
```
norrishd, line 129: Why don't we need to check `data['time'] >= segment_start` anymore/after all? (I trust ya, just curious)

seerpy/utils.py
```diff
@@ -60,11 +60,24 @@ def download_channel_data(data_q, download_function):
...
-        if 'int' in data_type:
+
+        column_names = channel_names
+
+        if meta_data['channelGroups.timestamped']:
+            # timestamped data is not stored in records as EDF data is
+            # it is just a sequence of (ts, ch1, ch2, ..., chN), (ts, ch1, ch2, ..., chN), ...
+            # the timestamp is milliseconds relative to chunk start
+            column_names = ['time'] + channel_names
+            data = data.reshape(-1, len(column_names))
```
seerwill, line 71: Does this line have any effect? i.e. if the data structure is already `[[ts, c1], [ts, c2], ...]` does reshaping do anything?

seerpy/utils.py
seerwill: Isn't this to do with how EDF encodes NaNs for float vs int? I feel like @dkeden knows lots about this

seerpy/utils.py
bjdoyle: The blocking into rows is only conceptual, it actually comes in as a 1D vector, this reshapes it to a 2D array.

seerpy/utils.py
bjdoyle: I was being extra cautious, but the excess data actually only appears at the end of the segment.

seerpy/utils.py
bjdoyle: Yeah I believe it's something to do with that but I don't actually have a definitive answer. This comment is a placeholder for that better explanation.

bjdoyle:
I am working on getting data samples from both the Sense and Siesta for test data, which I will put in a separate PR.

seerpy/utils.py
```diff
@@ -90,22 +101,32 @@ def download_channel_data(data_q, download_function):
...
+            data['time'] = (np.arange(data.shape[0]) *
+                            (1000.0 / meta_data['channelGroups.sampleRate'])
+                            + meta_data['dataChunks.time'])
+
         data['id'] = study_id
         data['channelGroups.id'] = channel_groups_id
         data['segments.id'] = segments_id
         data = data[['time', 'id', 'channelGroups.id', 'segments.id'] + channel_names]
 
         # data chunks seem to always be 10 seconds even if they don't contain that much data
```
robrkerr, line 127: It doesn't seem like you have (but just checking because of this comment), the chunk size should not be hardcoded - we typically use 10s but it can be anything.

seerpy/utils.py
```diff
@@ -90,22 +101,32 @@ def download_channel_data(data_q, download_function):
...
+        data[channel_names] = data[channel_names] * 10.0**exponent
+
         data = data.fillna(method='ffill', axis='columns')
         data = data.fillna(method='bfill', axis='columns')
         data = data.fillna(value=0., axis='columns')
-        data['time'] = (np.arange(data.shape[0]) * (1000.0 / meta_data['channelGroups.sampleRate'])
-                        + meta_data['dataChunks.time'])
+
+        if meta_data['channelGroups.timestamped']:
+            # data timestamp is relative to chunk start
```
robrkerr, line 114: You've probably checked but just in case not - we you sure it's the start of the chunk and not the start of the segment?

seerpy/utils.py
robrkerr: For the float encoding, missing values can be encoded using NaNs but this isn't an option for the int encodings. Instead we use the most negative integer value for this. It doesn't seem like the EDF specification has an explicit preference for this and people either use this approach of use 0 for missing values. The latter is obviously a terrible idea.

seerpy/utils.py
bjdoyle: Right, yeah we don't rely on what the chunk size is - but previously we would return the entire chunk, even if we were at the end of the segment and the chunk was padded to reach the chunk size. This just removes any padding.

seerpy/utils.py
bjdoyle: Yes.

seerpy/utils.py
bjdoyle: I'll update the comment in the next PR

seerpy/utils.py
bjdoyle: Cool, thanks. What about the dig_diff stuff - what's that about?

bjdoyle: Also, I get that this `data[np.all(data == np.iinfo(np.dtype(data_type)).min, axis=1), :] = np.nan` is replacing minimum int values with NaNs, but I'm not sure what the nan_mask_corrected stuff above that is trying to do?

bjdoyle: I think it is discarding any minimum ints at the end of the data, before it then converts any internal minimum ints to NaNs?

bjdoyle: So the dig_diff stuff is rebasing the data from min_int to max_int, into 0 to int_range, then converting that into a float value between 0 and 1, then rebasing that to be between signalMin and signalMax?

robrkerr: > What about the dig_diff stuff - what's that about?

That's for converting from the integer coding for sample values to a normal float encoding. Linear interpolation needs to be used to map from integers between digital min and digital max to the values they are encoding between signal min and signal max.

robrkerr: > I think it is discarding any minimum ints at the end of the data, before it then converts any internal minimum ints to NaNs?

Yeah I think so too.

robrkerr: > So the dig_diff stuff is rebasing the data from min_int to max_int, into 0 to int_range, then converting that into a float value between 0 and 1, then rebasing that to be between signalMin and signalMax?

Are you talking about the following line in the code?
> `data = (data - dig_min) / dig_diff * chan_diff + chan_min`

Yeah that's doing the linear interpolation of the sample values from their integer encoding to their actual values.

PR #92, opened 2020-07-15
Add test data and tests for utils.download_channel_data

bjdoyle:
Adds test data and tests for utils.download_channel_data.
Also makes sure that the time calculation is operating on 2 float64s - if the delta is a float32 it can cause problems.
Also added some comments and clarified others.


seerpy/utils.py
```diff
@@ -124,7 +130,8 @@ def download_channel_data(data_q, download_function):
         data['segments.id'] = segments_id
         data = data[['time', 'id', 'channelGroups.id', 'segments.id'] + channel_names]
 
-        # data chunks seem to always be 10 seconds even if they don't contain that much data
+        # chunks are padded to a standard size (usually 10s) if they don't contain that much data.
```
robrkerr, line 133:
```suggestion
        # chunks are padded to a chunk size specified for the channel group (usually 10s) if they don't contain that much data.
```

seerpy/utils.py
```diff
@@ -112,8 +117,9 @@ def download_channel_data(data_q, download_function):
 
         if meta_data['channelGroups.timestamped']:
             # data timestamp is relative to chunk start
-            chunk_start = meta_data['dataChunks.time']
-            data['time'] = data['time'] + chunk_start
+            # make sure both are float64 - sometimes mixed float arithmetic gives strange results
```
robrkerr, line 120: Yeah that makes sense given the precision we are interested.

tests/test_utils.py
```diff
@@ -15,6 +16,134 @@
...
+            'channelGroups.samplesPerRecord': [250],
+            'channelGroups.recordsPerChunk': [10],
+            'channelGroups.compression': ['gzip'],
+            'channelGroups.signalMin': [0],
+            'channelGroups.signalMax': [0],
+            'channelGroups.exponent': [-6],
+            'channelGroups.timestamped': [True]
+        }
+        meta_data = pd.DataFrame(meta_data_dict).iloc[0]
+        print('meta_data.dtypes', meta_data.dtypes)
```
robrkerr, line 47: Did you mean to leave in these print statements?

seerpy/utils.py
bjdoyle: I've reworded that

tests/test_utils.py
bjdoyle: No, removed.

PR #97, opened 2020-08-24
Add data downloader to Example scripts folder

dkeden:
Data downloader scripts for Epilepsy Ecosystem. 

The file to run is `download_data.py` which uses scripts in the folder `downloader`. 

May need to modify file & folder names. Unsure if main file should replace the file `contestant_data_downloader.py`?

@EwanNurse

### DIFF
--- a/Examples/download_data.py
+++ b/Examples/download_data.py
@@ -22,7 +22,8 @@ def get_download_status_file(client):
                 'fff9aaa9-b104-46e8-9227-b1b76d6f333e', '372482db-bcd8-4c79-85aa-16de76d8df69',
                 '92434006-5506-4d4a-97e4-f7232dfd68bd', '4f88cf00-8bf3-43fd-970d-9e434acf6edb',
                 'fba0f4f1-29c5-4387-b076-c6f788bbcaca', '72397e46-c424-4580-af50-1c460894664d',
-                'bab85744-4040-4496-b35e-dfb1ed638b73', 'ab1d9a55-190a-431f-ad31-653e137f584c'
+                'bab85744-4040-4496-b35e-dfb1ed638b73', 'ab1d9a55-190a-431f-ad31-653e137f584c',
+                '446364b3-ccb3-4927-8a32-f70ab0f5aa13', '3112c100-d71e-438f-96e5-4defc3b9cfa7'
             ]
         }
         write_json(studies_file, status)
--- a/Examples/download_data.py
+++ b/Examples/download_data.py
@@ -3,6 +3,7 @@ import sys
 import time
 import argparse
 
+from os import mkdir
 from os.path import dirname, isdir, isfile, join
 from seerpy import SeerConnect
 from urllib3.exceptions import ReadTimeoutError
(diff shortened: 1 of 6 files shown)


Examples/download_data.py
```diff
@@ -0,0 +1,99 @@
...
+                sys.stdout.write(
+                    f'ReadTimeoutError. Re-trying after a short break... {str(i)}'
+                )
+                sys.stdout.flush()
+                time.sleep(1)
+
+        # If user reaches 5 attempts, break and request user to return later
+        if attempts == 5:
+            print("ReadTimeoutError! Number of tries exceeded. Please re-run \
+                this script at a later time.")
```
bjdoyle, line 87:
```suggestion
            print("ReadTimeoutError! Number of tries exceeded. Please re-run this script at a later"
                  " time.")
```

Examples/downloader/downloader.py
```diff
@@ -0,0 +1,159 @@
...
+        self.label_groups = self.get_label_groups()
+        self.channel_groups = self.get_channel_groups()
+
+        self.study_name = self.channel_groups['name'][0]
+        self.folder_out = join(path_out, self.study_name)
+        if not isdir(self.folder_out):
+            mkdir(self.folder_out)
+
+    def get_channel_groups(self):
+        channel_groups = self.client.get_all_study_metadata_dataframe_by_ids(
```
bjdoyle, line 33:
```suggestion
        channel_groups = self.client.get_all_study_metadata_dataframe_by_ids([self.study_id])
```

bjdoyle: Actually could you run yapf over this?

Examples/downloader/downloader.py
```diff
@@ -0,0 +1,159 @@
...
+                'segments.timezone', 'channels.channelType.category',
+                'channels.channelType.name'
+            ])
+        return channel_groups
+
+    def get_label_groups(self):
+        return self.client.get_label_groups_for_studies([self.study_id])
+
+    def get_channels(self):
+        channels = {}
```
bjdoyle, line 47: Something like this:
```suggestion
        return self.channel_groups[['channelGroups.name', 'channels.name']].drop_duplicates()
```
I haven't tested this, it may not be exactly right. This will return a dataframe, you can convert it to a list using .to_numpy().tolist()

bjdoyle (changes requested):

bjdoyle (commented):
Did Levin request that the output be in json? It's an interesting choice.
Could you run yapf over this for me, it's mostly good but there's a few things that makes me think it hasn't been run.
And get_channel_groups is more complicated than it needs to be.
Otherwise looks good.

Examples/downloader/downloader.py
dkeden: Ah! Okay, I've just added in a change

dkeden:
> Did Levin request that the output be in json? It's an interesting choice.
> Could you run yapf over this for me, it's mostly good but there's a few things that makes me think it hasn't been run.
> And get_channel_groups is more complicated than it needs to be.
> Otherwise looks good.

He didn't, but he wasn't happy with the size of the CSV files or the time it took to write each file despite using speedy vectorization w/ numpy arrays and writing to h5 file formats first, so I suggested JSON - just 'cause it's light-weight, fairly universal, MATLAB-friendly and approved by Kaggle.

Awesome, thanks. 

I think I've run yapf. Does this version look better?

Examples/downloader/downloader.py
dkeden: Yup! I think I've done it this time?

EwanNurse:
Regarding naming, we'll need to keep `contest_data_downloader.py` just to keep supporting the NeuroVista download. Maybe rename to 'neurovista_contest_downloader.py' and add 'msg_' to the start of 'download_data.py'?

dkeden:
> Regarding naming, we'll need to keep `contest_data_downloader.py` just to keep supporting the NeuroVista download. Maybe rename to `neurovista_contest_downloader.py` and add 'msg_' to the start of `download_data.py`?

Sounds good, I'll make a change now

bjdoyle (changes requested):
get_channels could still do with a modification

Examples/downloader/downloader.py
```diff
@@ -0,0 +1,134 @@
...
+            'channels.channelType.name'
+        ])
+        return channel_groups
+
+    def get_label_groups(self):
+        return self.client.get_label_groups_for_studies([self.study_id])
+
+    def get_channels(self):
+        channels = self.channel_groups[['channelGroups.name', 'channels.name']].drop_duplicates()
+        return ([channel_group, channel] for channel_group, channel in zip(
```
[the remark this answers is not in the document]
dkeden: Cool, didn't realise you could do this

Examples/downloader/downloader.py
```diff
@@ -43,7 +43,7 @@ def get_label_groups(self):
 
     def get_channels(self):
         channels = self.channel_groups[['channelGroups.name', 'channels.name']].drop_duplicates()
-        return ([channel_group, channel] for channel_group, channel in zip(
+        return channels.to_numpy().tolist()
             channels['channelGroups.name'].tolist(), channels['channels.name'].tolist()))
```
bjdoyle, line 47: But you also have to delete this line! (Sorry, I can't figure out how to get suggest change to encompass multiple lines)

bjdoyle (changes requested):

Examples/downloader/downloader.py
dkeden: Oh... sorry I was in speedy mode this morning

dkeden: Should be good now

bjdoyle (approved):

PR #100, opened 2020-08-25
Update README.md

EwanNurse:
Update text to include MSG data downloader instructions


README.md
```diff
@@ -8,7 +8,15 @@ To install, simply clone or download this repository, then type `pip install .`
 
 ### Epilepsy Ecosystem Data
 
-For users attempting to download data for the [Epilepsy Ecosystem](https://www.epilepsyecosystem.org/howitworks/), please download the [latest release](https://github.com/seermedical/seer-py/releases/latest) instead of cloning the repository or downloading the master branch. Then open the script `ContestDataDownloader.py` in `Examples` and it will guide you through the download process (you will need to change a few things in this script including the path to download the data to).
+For users attempting to download data for the [Epilepsy Ecosystem](https://www.epilepsyecosystem.org/howitworks/), please download the [latest master version](https://github.com/seermedical/seer-py/tree/master).
+
+#### My Seizure Gauge Data
+
+For the My Seizure Gauge dataset, open the script `msg_data_downloader.py` in `Examples` and it will guide you through the download process (you will be prompted in the command line for the path to download the data to).
```
dkeden, line 15: Great! At the moment I've added an argparse argument, so that when the user runs the script they'd type something like `python run.py -p /Users/username/mydata` so that if the script is aborted, etc., the user can just re-run the same line w/out having to specify the path again. 

In saying that, I can easily add in a keyword arg so that the user is prompted to change their path from a default path?

EwanNurse: Thank you for the clarification! Updated, and removed Examples/download_data.py

dkeden: Awesome!

dkeden: Can we safely assume that all (incl. MATLAB users) would know how to run a script from their terminal? Otherwise it might be worth mentioning,

> run the script by navigating to the directory where XX is located, in your terminal. Run python msg_data_downloader.py -o /path/to/directory/ etc.

dkeden: If not, all good!

dkeden (approved):

README.md
EwanNurse: I take your point, but this is a Python API, we do need to have some level of expected knowledge.

dkeden: Good

PR #102, opened 2020-08-26
Small bug fix for msg downloader

dkeden:
This PR removes the dropping of columns from the channel group metadata dataframe, as some data do not contain these columns.


EwanNurse (approved):
This change will mean some extra blank fields are saved in the downloader, which is not ideal, but will not change the overall functionality (besides marginally increasing the final file size). For some reason using the `error = 'ignore'` flag in `.drop` didn't work in solving this problem, due to an error seerpy was throwing. I recommend when we revisit the NeuroVista downloader to revisit this as well.

PR #103, opened 2020-09-01
Add queryName and query variables to queries used by seer-worker

dkeden:
This PR adds query names and variables


seerpy/graphql.py
```diff
@@ -435,57 +436,68 @@ def get_diary_created_at_query_string(patient_id):
...
+            $value: String!,
+            $limit: PaginationAmount, 
+            $offset: Int,
+            $from_time: Float!,
+            $to_time: Float!,
+            $from_duration: Float!,
+            $to_duration: Float!) {
+                    patient (id: $id) {
                         id
-                        labelType
```
dkeden, line 449: not sure what has happened here, will have to fix up

seerpy/graphql.py
```diff
@@ -388,7 +388,7 @@ def get_user_from_patient_query_string(patient_id):
 
 def get_patients_query_string():
     return """
-        query {
+        query getPatientList {
```
bjdoyle, line 391: ~~I think it's better to actually convert this query to use variables or remove this name~~
Sorry, got confused about line 436 being part of this function.

bjdoyle (approved):
Maybe add a comment to PR that lastInsightGenerated is added and add_user_timezone is removed.
Otherwise, assuming it all works, then I'm happy.

PR #104, opened 2020-09-05
Add pre-commit config file

norrishd:
Should we add a [pre-commit](https://pre-commit.com/) config file so that code quality checks can be run on each commit? And instructions in the "Development" section of the README.md for how to enable it? I know there's no way to enforce that developers do enable it, so it's not as robust a safety measure as Buildkite, but I personally still find it super helpful!

Currently hooks set up (these might not be the optimal set):
- Run YAPF
- Check any Jupyter notebook files added don't have output included (but you can whitelist exceptions, e.g. the Example notebook).
- Check no large files accidentally committed
- Check AWS credentials weren't somehow accidentally committed
- Check no debug statements left in code

And on every push:
- Run the tests

P.s. Ignore all files changes except to:
- .pre-commit-config.yaml
- README.md

The other changes are YAPF running as a pre-commit hook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,4 +122,5 @@ To set up pre-commit (highly recommended):
     pre-commit install
     pre-commit install -t pre-push
 
-Now code will automatically be linted and formatter in various ways before every commit, and tests run before every push.
+Now code will automatically be linted and formatted in various ways before every commit, and tests run before every push.
+
--- a/README.md
+++ b/README.md
@@ -115,3 +115,11 @@ There is a known issue with using python's multiprocessing module on Windows wit
 1. To format the code using yapf, run `yapf -ir seerpy tests`
 2. To run pylint on the code, run `pylint seerpy tests`
 3. To run tests and generate an html coverage report, run `pytest --cov-report=html --cov=seerpy`
+
+To set up pre-commit (highly recommended):
+
+    pip install pre-commit  # see https://pre-commit.com/ for alternatives
+    pre-commit install
+    pre-commit install -t pre-push
+
+Now code will automatically be linted and formatter in various ways before every commit, and tests run before every push.
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # seer-py
 
-Python SDK for the Seer data platform, which handles authenticating a user, downloading channel data, and uploading labels/annotations.
+Python SDK for the Seer data platform, which handles authenticating a user, downloading channel or other data, and uploading labels/annotations.
 
 ## Install
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,12 +18,15 @@ repos:
         exclude: Examples/Example.ipynb
 -   repo: local
     hooks:
-    -   id: tests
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -115,3 +115,11 @@ There is a known issue with using python's multiprocessing module on Windows wit
...
 2. To run pylint on the code, run `pylint seerpy tests`
 3. To run tests and generate an html coverage report, run `pytest --cov-report=html --cov=seerpy`
+
+To set up pre-commit (highly recommended):
+
+    pip install pre-commit  # see https://pre-commit.com/ for alternatives
+    pre-commit install
+    pre-commit install -t pre-push
+
+Now code will automatically be linted and formatter in various ways before every commit, and tests run before every push.
```
seerwill, line 125:
```suggestion
Now code will automatically be linted and formatted in various ways before every commit, and tests run before every push.
```

seerwill (commented):
Nice idea! I'm nominating myself as an extra reviewer 😁 

Some thoughts:

- Running tests before pulls seems an interesting choice - I'm not sure what problem this addresses given the code has already been committed and tests are being run in CI anyway? It seems this would prevent pushing up WIP on larger PRs (without e.g. commenting out or removing failing tests).
- Does the `repo` item for `mirrors-yapf` pull a separate copy of yapf or is this just a "runner" script? If it is separate, how can we ensure that it formats in the same way that CI expects?

norrishd:
Thanks for the comments @seerwill! Re. YAPF, I might be wrong about this, but I'm pretty sure it does pull a separate copy of YAPF, and that copy then discovers and uses the style sheet in the repo. Yeah that could present a problem if CI was formatting differently.

And good point about the tests! I guess it's a philosophical thing whether you want to be able push falling tests. For myself I'd rather know right away that I've broken something, and then have to `pytest.mark.xfail()` before pushing them, so I don't forget and if I do at least reviewers can clearly see. Because we don't have Buildkite running automatically on seer-py, I know I've opened a PR that broke a test (I think I might've reviewed one of yours that did too 😜)

bjdoyle:
How long does it take to run, and does it lint / format everything or just what has been changed?

norrishd:
By default, [pre-commit only runs against changed files](https://pre-commit.com/#4-optional-run-against-all-the-files), which is why the "run tests" hook needs `always_run: true` to be specified; otherwise it would only run test files that have changed. (`stages: [push]` makes that hook only run when pushing).

It took 1.4 seconds just now to manually invoke on all files with `pre-commit run -a`, and most of that time was YAPF, so it would typically be a little faster only running on changed files. Doing a push, i.e. so it also runs the tests, took 4.2 seconds, because our tests are so fast and lean.

One extra little annoyance/consideration if we turned on the pre-push test runner is it'll try to call `pytest` using whatever virtual env is active when you `git push`, so have to make sure it's one with `pytest`/`jwt` etc installed

bjdoyle:
That seems fast enough. I'll give it a whirl and see what it feels like. But in general it seems like a good idea, as long as it's fast enough to not be annoying.

PR #105, opened 2020-09-07
Add keys to query response in get_diary_labels()

dkeden:
It seems I may have missed the keys for `patient` and `diary` in `get_diary_labels()` response from API. Sorry about that.


seerwill (approved):
LGTM

PR #106, opened 2020-09-21
Add edit label group mutation

bjdoyle:
Adds a new mutation to edit a study label group (name, description).
Adds description field to get label groups query (so you can set the original in edit label group).
Adds a test file to check parsing of graphql queries.

### DIFF
--- a/tests/test_graphql.py
+++ b/tests/test_graphql.py
@@ -3,7 +3,7 @@ from gql import gql
 import seerpy.graphql as graphql
 
 
-def test_graphql_query_string():
+def test_graphql_query_strings():
     """Ensure query strings parse correctly."""
     gql(graphql.get_add_labels_mutation_string())
     gql(graphql.get_tag_id_query_string())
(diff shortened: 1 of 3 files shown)


tests/test_graphql.py
```diff
@@ -0,0 +1,16 @@
...
+    """Ensure query strings parse correctly."""
+    gql(graphql.get_add_labels_mutation_string())
+    gql(graphql.get_tag_id_query_string())
+    gql(graphql.EDIT_STUDY_LABEL_GROUP)
+    gql(graphql.get_organisations_query_string())
+    gql(graphql.get_patients_query_string())
+    gql(graphql.get_diary_labels_query_string())
+    gql(graphql.get_diary_medication_alerts_query_string())
+    gql(graphql.get_diary_medication_alert_windows_query_string())
+    gql(graphql.get_diary_medication_compliance_query_string())
```
bjdoyle, line 16: I have another PR which changes these all to variables instead of functions and converts more queries to use query variables

seerwill (approved):
Looks good, with caveats discussed on slack! I'd only recommend renaming the function to avoid confusion with diary labels.

seerwill (approved):

PR #108, opened 2020-10-02
Clarify API key string implementation

pkaroly:
Update the README with an example of how to pass an API key as a string (i.e. for AWS or environments where a pem key file should not be stored)


bjdoyle (approved):
👍

PR #109, opened 2020-10-02
update README to clarify how to get data access

bjdoyle:
Also updated the method reference in the note about windows multiprocessing, but that whole note should be revisited to check if it's still required - could a windows folk confirm?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,9 @@
 
 Python SDK for the Seer data platform, which handles authenticating a user, downloading channel data, and uploading labels/annotations.
 
+## Access to data
+Note that by default there is no public access to any data on the Seer Cloud. You need to be added to an organisation within the Seer Cloud in order to gain access to that organisation's data. The only publicly accessible data is through the https://www.epilepsyecosystem.org/ organisation.
+
 ## Install
 
 To install, simply clone or download this repository, then type `pip install .` which will install all the dependencies.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,9 @@
 
 Python SDK for the Seer data platform, which handles authenticating a user, downloading channel data, and uploading labels/annotations.
 
+## Access to data
+Note that by default there is no public access to any data on the Seer Cloud. You need to be added to an organisation within the Seer Cloud in order to gain access to that organisation's data. The only publicly accessible data is through the https://www.epilepsyecosystem.org/ organisation.
```
EwanNurse, line 6: My understanding of Epilepsy Ecosystem is that the users are actually part of a User Group, not specific organisations. Can I suggest the wording be:

> Note that by default there is no public access to any data on the Seer Cloud. You need to be added to an organisation or user group within the Seer Cloud in order to gain access to data. The only publicly accessible data is through the https://www.epilepsyecosystem.org/ user group.

EwanNurse (changes requested):

README.md
bjdoyle, line 6:
```suggestion
Note that by default there is no public access to any data on the Seer Cloud. You need to be added to an organisation or user group within the Seer Cloud in order to gain access to data. The only publicly accessible data is through the https://www.epilepsyecosystem.org/ user group.
```

EwanNurse (approved):
Looks good. Can confirm the download hangs in Spyder as described, and that suggested methods resolve this on my local Windows machine.

PR #110, opened 2020-10-09
Convert more queries to use graphql variables

bjdoyle:
Converts more queries to use graphql variables rather than python string formatting.
In particular, all queries which use get_paginated_response have been converted.
Also adds test utility methods which make testing queries which use get_paginated_response easier, cutting out a lot of boilerplate.

Also includes a change to the query string for get_diary_insights as the existing one was found to not be consistent with the current schema.

As well as test methods, all changed queries have been tested live against the api, with the exception of get_study_ids_in_study_cohort and get_user_ids_in_user_cohort, for which I could not find suitable data to test against.

Note that these tests were to ensure that the queries run without error. It would be appreciated, if you use particular queries, if you could confirm that they still perform as expected, in particular that they are returning data in the structure that you expect.

Further improvements to testing are required, in particular that the interpolation of variables is not tested in the current tests as that happens when the query is run, which is currently mocked. I have not fixed that in this PR, I am still thinking of a way to do it. It would also be good to have integration tests which run the queries against actual data but again that is a job for the future.


tests/test_seerpy.py
```diff
@@ -66,6 +66,200 @@ def test_query_variables_are_passed_on_initial_failure(self, gql_client, unused_
...
+    def test_get_documents_for_studies(self, gql_client, unused_sleep, seer_connect):
+        # run test and check result
+        self.check_paginated_query_with_data_variations(
+            gql_client, seer_connect, function_to_test=seer_connect.get_documents_for_studies,
+            function_args=[['study-1-id', 'study-2-id']], response_file='study_documents.json',
+            empty_response={'studies': []})
+
+    def test_get_diary_insights(self, gql_client, unused_sleep, seer_connect):
+        # TODO: I'm uncertain if the values in this file are appropriate
+        # TODO: There is a mis-match between the fields in the query and the schema in graphiql
```
seerwill, line 199: The schema is the source of truth. The API has been evolving and we haven't gotten around to updating seerpy yet.

tests/test_seerpy.py
```diff
@@ -66,6 +66,200 @@ def test_query_variables_are_passed_on_initial_failure(self, gql_client, unused_
         assert gql_client.return_value.execute.call_args[1]['variable_values'] == {'a': 'b'}
 
 
+@mock.patch('time.sleep', return_value=None)
+@mock.patch('seerpy.seerpy.GQLClient', autospec=True)
+class TestPaginatedQuery:
+    @classmethod
+    def check_paginated_query(cls, seer_connect, function_to_test, function_args=None,
```
seerwill, line 73: Nice work, this makes things a lot cleaner on the testing side!

seerwill (changes requested):
Looks good on the whole. There are some instances where optional fields are marked as required which I have noted. I think in most cases this will be OK as going this way won't be an error (marking a required field as optional will cause a schema validation error). However I think for the most part we should accept omitted fields for things like `fromTime` / `toTime` to prevent the need for dummy values like `90000000000`. 

There are also a few areas I flagged where using `Range` rather than separate from/to times would make the query much easier to use and more flexible. Would require a bit of rewriting the functions in `seerpy.py` so probably beyond the scope of this PR.

I agree integration testing would be nice here as its hard to tell just from looking at the schema and the JSON test files whether the queries will fail against the real API.

seerpy/graphql.py
```diff
@@ -810,36 +791,35 @@ def remove_studies_from_study_cohort_mutation_string(study_cohort_id, study_ids)
...
-        query {{{{
-            userCohort(id: "{user_cohort_id}") {{{{
+GET_USER_IDS_IN_USER_COHORT_PAGED = """
+    query userCohort($user_cohort_id: String,
+                     $limit: PaginationAmount,
+                     $offset: Int) {
+        userCohort(id: $user_cohort_id) {
+            id
+            name
+            users(limit: $limit, offset: $offset) {
```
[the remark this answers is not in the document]
bjdoyle: Wouldn't that change as is fail? I'd rather this go in a separate PR if it's required - this is only meant as a conversion to using variables as far as possible.

bjdoyle:
@seerwill I've accepted some of your suggestions, made some changes in response to others. I don't want to make changes in this PR that go beyond converting existing queries as-is to use query variables, as far as that has been possible, so some others I have not changed.

seerpy/graphql.py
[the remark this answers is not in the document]
seerwill: I don't think it would fail, but I could be wrong I just edited through github. I think it's more that the current format suggests you can omit `$user_cohort_id` which is not true. 

Happy for this to be in a separate PR I don't think its urgent.

seerwill:
> I don't want to make changes in this PR that go beyond converting existing queries as-is to use query variables

Makes sense, I think this PR has just highlighted some of the odd parts of the code base, so I though it was worth noting.

seerpy/graphql.py
```diff
@@ -446,10 +446,10 @@ def get_diary_created_at_query_string(patient_id):
...
                          $limit: PaginationAmount,
                          $offset: Int,
-                         $from_time: Float!,
-                         $to_time: Float!,
-                         $from_duration: Float!,
-                         $to_duration: Float!) {
+                         $from_time: Float,
+                         $to_time: Float,
+                         $from_duration: Float,
+                         $to_duration: Float) {
```
seerwill, line 452: I suspect this change will error out. The type in the schema for ranges is `Float!`, it's the `Range` itself that is optional. As the query hard codes `labels(ranges:...` then the floats are required to be populated.

For the purposes of this PR as you aren't planning to construct ranges in the python code, then these should be left as mandatory.

seerpy/graphql.py
```diff
@@ -489,8 +489,8 @@ def get_diary_created_at_query_string(patient_id):
 
 GET_DIARY_MEDICATION_ALERTS = """
     query getDiaryAlertMedication($patient_id: String!,
-                                  $from_time: Float!,
-                                  $to_time: Float!) {
+                                  $from_time: Float,
+                                  $to_time: Float) {
```
seerwill, line 493: Similar here, `Range` is optional, these are are required.

seerpy/graphql.py
```diff
@@ -543,8 +543,8 @@ def get_diary_created_at_query_string(patient_id):
 
 GET_DIARY_MEDICATION_COMPLIANCE = """
     query getMedicationCompliance($patient_id: String!,
-                                  $from_time: Float!,
-                                  $to_time: Float!) {
+                                  $from_time: Float,
+                                  $to_time: Float) {
```
seerwill, line 547: Same as above

seerwill (changes requested):
Commit 61f88c39981b386cdd5d53fa11c1fb2a29cb756a will have to be reverted I believe, but other than that looks basically good to go.

P.S. I think this is a good example of code that will pass unit testing but fail integration tests when they are written.

seerwill (approved):
👍

PR #111, opened 2020-10-11
Add patient ID to mood survey query

dkeden:
Small PR to add patient ID to the mood survey query


seerwill (dismissed):
Looks fine to me, might be worth coordinate with @bjdoyle around #110  before merging.

bjdoyle:
@dkeden I've merged my changes from the last PR into this. You might want to check you're happy. It would also be good if you could update the test files with this change (for completeness - though the tests pass as is).

seerwill (dismissed):

dkeden:
Awesome, thanks @bjdoyle. I've just updated the mood survey test files. I don't think there's anything else I need to do?

seerwill (approved):

bjdoyle (changes requested):
Need to change the result csvs to reflect the json (I think you already saw this)

bjdoyle (changes requested):
Sorry, I see you already did, but there is still a test error

seerwill (dismissed):
👍

PR #112, opened 2020-10-14
Enable the use of datachunk urls (including s3 urls) in get_channel_data

bjdoyle:
Changes get_channel_data to allow the use of data chunk urls (including s3 urls) as well as the existing segment urls.
Also fixes some errors in get_data_chunk_urls and renames the data chunk columns to mirror the existing datachunk columns from utils.create_data_chunk_urls. (I'm pretty sure no-one was previously using seerpy.get_data_chunk_urls or they would have noticed it's problems).

The default if you call get_channel_data with default parameters will now be to use s3 to download instead of cloudfront, which is cheaper for us as long as the retrieving code is running in the same aws region as the data, and otherwise is equivalent.

### DIFF
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -660,15 +660,16 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
     def get_data_chunk_urls(self, study_metadata, s3_urls=True, from_time=0, to_time=9e12,
                             limit=10000):
         """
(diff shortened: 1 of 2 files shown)


robrkerr (commented):
Looks good to me but I'll let someone else review it in more detail.

seerwill (approved):
Looks fine to me, although I'm a little hazy on the details of the chunk URLs.

PR #114, opened 2020-10-19
Add de and us regions to auth

bjdoyle:
Adds the de and us regions to the list of valid regions able to be automatically detected from api key file names.


norrishd (approved):
Ohhh yeah

PR #115, opened 2020-10-21
Filter diary labels by tagType value

dkeden:
This PR modifies the strategy to filter diary labels. 

Previously, the `get_diary_labels` query used an API-based filter on `labelGroups` (i.e. filters by `labelType`). Unfortunately, `labelGroups` are based on a question when a user creates an account, rather than the labels themselves. In order to retrieved all labels of a specific type, e.g. "Seizure", "Cardiac" or "Event", the only current way is to manually filter by `tagType` value. 

It would be ideal to modify / add a filter in the API to filter this instead.

P.s. Haven't compelted tests yet, and will need to finish other mood surveys PR first.


tests/test_seerpy.py
```diff
@@ -709,6 +709,60 @@ def test_generating_remove_studies_from_cohort_mutation(self):
...
+    def test_get_diary_labels_no_matching_labels(self, gql_client, unused_sleep, seer_connect):
+        # setup
+        side_effects = []
+
+        with open(TEST_DATA_DIR / "diary_labels_2.json", "r") as f:
+            side_effects.append(json.load(f))
+
+        gql_client.return_value.execute.side_effect = side_effects
+
+        expected_result = {'labelGroups': [{'numberOfLabels': 1, 'labels': []}]}
```
dkeden, line 759: I have a small issue here --

Being that some labels may match the filter criteria, e.g. "Seizure" and others may not, the `numberOfLabels` from the API won't reflect the 'filtered' number of labels.

Can either leave as is (as I believe this is a temporary work-around), or reconstruct the dictionary, or replace the `numberOfLabels` field at end of interation/s.

Thoughts? @bjdoyle @seerwill

seerwill: I think if you are filtering the response from the server you should also rebuild the `numberOfLabels` to match.

bjdoyle: I think it is the total number of labels in the group, so it will be the same on each page. I don't think it should be modified, if anything I would drop it from the query. But you'd have to look at the api to be sure.

seerwill: Dropping it works too, although I don't know where this field is being used via seerpy (if at all).

dkeden: Just returns the page size (if that means just returns the total number of labels that exist)

dkeden: So by modifying the `numberOfLabels` in the seerpy response, we'd technically be changing the definition

dkeden: Maybe?

dkeden: I've updated the `numberOfLabels`

seerpy/seerpy.py
```diff
@@ -1173,29 +1172,36 @@ def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, fr
                 labels = group['labels']
 
                 if not labels:
+                    label_results = response
                     continue
 
+                filtered_labels = []
+                for label in labels:
+                    if any(tag['tagType']['value'] == label_type.capitalize() for tag in label['tags']):
```
bjdoyle, line 1180: it would be better to do this the other way around and take the lower of the tag.

seerwill: This PR has been closed in favour of #117

PR #117, opened 2020-10-22
Filter diary labels by tagType value

dkeden:
This PR modifies the strategy to filter diary labels. 

Previously, the `get_diary_labels` query used an API-based filter on `labelGroups` (i.e. filters by `labelType`). Unfortunately, `labelGroups` are based on a question when a user creates an account, rather than the labels themselves. In order to retrieved all labels of a specific type, e.g. "Seizure", "Cardiac" or "Event", the only current way is to manually filter by `tagType` value. 

It would be ideal to modify / add a filter in the API to filter this instead.

Second attempt.


seerpy/seerpy.py
```diff
@@ -1111,7 +1111,7 @@ def get_diary_created_at(self, patient_id):
         response = self.execute_query(query_string)
         return response['patient']['diary']['createdAt']
 
-    def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, from_time=0,
+    def get_diary_labels(self, patient_id, label_type='all', tag_type='seizure', offset=0, limit=100, from_time=0,
```
seerwill, line 1114: Have you confirmed that `label_type='all'` works as expected?

seerpy/seerpy.py
```diff
@@ -1121,8 +1121,11 @@ def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, fr
         patient_id : str
             The patient ID for which to retrieve diary labels
         label_type : str, optional
-            The type of label to retrieve. Default = 'all'. Options = 'seizure',
+            The type of label groups to retrieve. Default = 'all'. Options = 'seizure',
             'medications', 'cardiac'.
+        tag_type : str, optional
+            The type of label to retrieve. Default = 'seizure'. Options = 'seizure',
+            'cardiac', 'other'.
```
seerwill, line 1128:
```suggestion
            The type of label to retrieve. Default = 'seizure'. Options = 'seizure',
            'cardiac', 'other'. Filters client-side using label tags.
```

seerpy/seerpy.py
```diff
@@ -1173,14 +1176,22 @@ def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, fr
                 labels = group['labels']
 
                 if not labels:
+                    label_results = response
```
seerwill, line 1179: Wouldn't this have undesirable results if for instance the first labelGroup has labels, but the second labelGroup doesn't? i.e. wouldn't it "reset" the `label_results` variable?

```suggestion
                    continue
```

seerpy/seerpy.py
```diff
@@ -1173,14 +1176,22 @@ def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, fr
...
                     continue
 
+                filtered_labels = []
+                for label in labels:
+                    if any(tag['tagType']['value'] == tag_type.capitalize() for tag in label['tags']):
```
seerwill, line 1184: @bjdoyle 's comment from #115

> it would be better to do this the other way around and take the lower of the tag.

seerpy/seerpy.py
```diff
@@ -1189,13 +1200,14 @@ def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, fr
...
-                label_results['labelGroups'][idx]['labels'].extend(labels)
+                label_results['labelGroups'][idx]['labels'].extend(filtered_labels)
+                label_results['labelGroups'][idx]['numberOfLabels'] += len(filtered_labels)
 
             offset += limit
 
         return label_results
 
-    def get_diary_labels_dataframe(self, patient_id, label_type='all', offset=0, limit=100,
+    def get_diary_labels_dataframe(self, patient_id, label_type='seizure', offset=0, limit=100,
```
seerwill, line 1210: I think its a bit confusing to have different defaults on these two queries. I'd set the default on both to `all`?

tests/test_seerpy.py
```diff
@@ -850,3 +850,58 @@ def test_empty(self, gql_client, unused_sleep, seer_connect):
 
         # check result
         assert result == expected_result
+
+
+@mock.patch('time.sleep', return_value=None)
+@mock.patch('seerpy.seerpy.GQLClient', autospec=True)
+class TestGetDiaryLabels:
```
seerwill, line 857: Is there a way to integrate these tests with the pagination approach used above?

seerwill (changes requested):
There don't appear to be any tests checking filtering across multiple label groups? It might be worth adding something along these lines.

seerpy/seerpy.py
dkeden: Thanks for pointing this out. Do you mean between between `get_diary_labels` `get_diary_labels_dataframe`? My preference would be to have both `label_type=all` and `tag_type=seizure` in both functions so that we don't break pre-existing code (as you mentioned) for filtering on `label_type`

tests/test_seerpy.py
dkeden: We don't actually paginate the diary labels, it's more of a 'manual' method, so not sure if we can? Or do you mean to develop a test that tests the fake pagination?

seerpy/seerpy.py
dkeden: Not other than in graphql, but I'd imagine that this would (or, should) be an API test? I've seen both "seizures" and "medications" `labelGroups` returned, etc.

seerpy/seerpy.py
```diff
@@ -1174,13 +1177,21 @@ def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, fr
 
                 if not labels:
                     continue
+                    continue
```
dkeden: Yeah, good point

dkeden: I guess it means we may end up with an empty dict of "medication" `labelGroups`, but I don't think that's a big issue if we'll change this in the API at some point

seerpy/seerpy.py
seerwill: Yes, I think using `all` is the best approach.

seerpy/seerpy.py
```diff
@@ -1172,15 +1175,21 @@ def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, fr
...
                 labels = group['labels']
 
-                if not labels:
-                    continue
+                filtered_labels = []
+                if labels:
+                    for label in labels:
+                        if any(tag['tagType']['value'].lower() == tag_type
+                            for tag in label['tags']):
+                            filtered_labels.append(label)
```
dkeden, line 1183: @seerwill I've modified this so that `[]` isn't returned, and the outer structure / response with an empty field on `labels` is returned. A test to cover this would be good at some point, I can put it on my TODO?

seerpy/seerpy.py
```diff
@@ -1188,14 +1188,24 @@ def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, fr
...
 
+        # filter
+        for idx, group in enumerate(label_results['labelGroups']):
+            filtered_labels = []
+            labels = group['labels']
+            if labels:
+                for label in labels:
+                    if any(tag['tagType']['value'].lower() == tag_type
+                        for tag in label['tags']):
+                        filtered_labels.append(label)
```
seerwill, line 1203:
```suggestion
                            filtered_labels.append(label)
```

Missing indentation?

seerpy/seerpy.py
```diff
@@ -1188,14 +1188,24 @@ def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, fr
...
+                label_results['labelGroups'][idx]['labels'] = filtered_labels
+                label_results['labelGroups'][idx]['numberOfLabels'] = len(filtered_labels)
         return label_results
 
-    def get_diary_labels_dataframe(self, patient_id, label_type='all', offset=0, limit=100,
+    def get_diary_labels_dataframe(self, patient_id, label_type='all', tag_type='seizure', offset=0, limit=100,
```
seerwill, line 1208: As above - this changes the default behaviour. It would be good to have an "all" option for tag_types

seerpy/seerpy.py
```diff
@@ -1188,14 +1188,24 @@ def get_diary_labels(self, patient_id, label_type='all', offset=0, limit=100, fr
...
+        for idx, group in enumerate(label_results['labelGroups']):
+            filtered_labels = []
+            labels = group['labels']
+            if labels:
+                for label in labels:
+                    if any(tag['tagType']['value'].lower() == tag_type
+                        for tag in label['tags']):
+                        filtered_labels.append(label)
+                label_results['labelGroups'][idx]['labels'] = filtered_labels
+                label_results['labelGroups'][idx]['numberOfLabels'] = len(filtered_labels)
```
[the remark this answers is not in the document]
bjdoyle: limit is the page size you shouldn't have to worry about it.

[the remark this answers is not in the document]
bjdoyle: but I don't think it matters if you filter as you go or at the end.

tests/test_data/diary_labels_2.json
```diff
@@ -0,0 +1,31 @@
...
+                            "timezone": 10,
+                            "tags": [
+                                {
+                                "tagType": {
+                                    "value": "Mobile app"
+                                    }
+                                },
+                                {
+                                "tagType": {
+                                    "value": "Cardiac"
```
seerwill, line 21: As this is a `cardiac` label, we don't know if the second request is merged in to the first with the `test_get_diary_labels` test below. It would be better to mark this as seizure and check for two labels in the result?

tests/test_seerpy.py
seerwill, line 857: I don't see a test here for merging multiple labelGroups (unless this is covered by existing tests somewhere?). As this is a more important aspect of the query now I think this should be covered by testing. 

(Sorry, I thought I'd commented on this previously but I can't seem to find the comment any more 🤷‍♂️)

seerwill (changes requested):
Still a few issues with tests and defaults on queries I think need to be resolved before merging.

PR #121, opened 2020-11-30
Add script to download studies in Examples folder

dkeden:
This is a modified version of `msg_data_downloader.py` in the Examples folder. By default, all studies, channel groups and channels are downloaded, unless a specific channel group is specified on input.

This script could be developed if there is a need for it in the future.

### DIFF
--- a/Examples/download_studies.py
+++ b/Examples/download_studies.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+import sys
+import time
+import argparse
+
+from os import mkdir
+from os.path import dirname, isdir, isfile, join
+from urllib3.exceptions import ReadTimeoutError
+from downloader.utils import read_json, write_json
+from seerpy import SeerConnect
+from downloader.downloader import DataDownloader
+
+
+def get_download_status_file(client):
+    """Retrieves JSON of download status for all study IDs."""
+    studies_file = join(dirname(__file__), 'studies.json')
+    if not isfile(studies_file):
+        status = {
+            study_id: 0
+            for study_id in client.get_study_ids()
+        }
+        write_json(studies_file, status)
+    return studies_file
+
+
+def run(client, path_out, channel_group_to_download):
+    """Downloads all data available on Seer's public API."""
+    if path_out:
+        if not isdir(path_out):
+            print('Please specify a path that exists.')
+            return
+        output_dir = join(path_out, 'data')
+    else:
+        output_dir = join(dirname(__file__), 'data')
+    if not isdir(output_dir):
+        mkdir(output_dir)
+
+    download_status_file = get_download_status_file(client)
+    download_status = read_json(download_status_file)
+    study_ids = [study_id for study_id in download_status if download_status[study_id] == 0]
+
+    attempts = 0
+    while True:
+        try:
+            if attempts != 0:
+                time.sleep(3)
+                client = SeerConnect()
+            for study_id in study_ids:
+                downloader = DataDownloader(client, study_id, output_dir)
+                # Download channel data
+                downloader.download_channel_data(channel_group_to_download)
+                download_status[study_id] = 1
+                write_json(download_status_file, download_status)
+            print('Download completed.')
+            return
+
+        except ReadTimeoutError:
+            attempts += 1
+            for i in range(5, 0, -1):
+                sys.stdout.write(f'ReadTimeoutError. Re-trying after a short break... {str(i)}')
+                sys.stdout.flush()
+                time.sleep(1)
+
+        # If user reaches 5 attempts, break and request user to return later
+        if attempts == 5:
+            print("ReadTimeoutError! Number of tries exceeded. Please re-run \
+                this script at a later time.")
+            break
+
+    return
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-o", "--outpath", help="Path for data to be saved (optional).")
+    parser.add_argument("-cg", "--channelgroup", help="Name of channel group to download.")
+    args = parser.parse_args()
+
+    run(SeerConnect(), args.outpath, args.channelgroup)
--- a/Examples/download_studies.py
+++ b/Examples/download_studies.py
@@ -87,5 +87,5 @@ if __name__ == '__main__':
     parser.add_argument("-c", "--channels", nargs="+", help="Name of channels to download (optional).")
     parser.add_argument("-p", "--party_id", help="Party ID for specific organisation to download studies from.")
     args = parser.parse_args()
-    breakpoint()
+ 
     run(SeerConnect(), args.outpath, args.update, args.channelgroups, args.channels, args.party_id)
--- a/Examples/download_studies.py
+++ b/Examples/download_studies.py
@@ -39,34 +39,33 @@ def run(client, path_out, channel_group_to_download):
     download_status = read_json(download_status_file)
     study_ids = [study_id for study_id in download_status if download_status[study_id] == 0]
 
-    attempts = 0
-    while True:
-        try:
-            if attempts != 0:
-                time.sleep(3)
-                client = SeerConnect()
-            for study_id in study_ids:
+    for study_id in study_ids:
+        attempts = 0
+        while True:
+            try:
+                if attempts != 0:
+                    time.sleep(3)
+                    client = SeerConnect()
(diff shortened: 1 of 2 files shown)


Examples/downloader/downloader.py
```diff
@@ -64,10 +64,13 @@ def get_segment_data(self, folder_out, channel_group, channel, channel_metadata,
...
             yield data, segment_file
 
-    def download_channel_data(self):
+    def download_channel_data(self, channel_group_to_download=None):
         print(f'Downloading channel data for {self.study_name}...')
 
         for (channel_group, channel) in self.get_channels():
+            if channel_group_to_download:
+                if not channel_group == channel_group_to_download:
+                    continue
```
bjdoyle, line 73:
```suggestion
            if channel_group_to_download and channel_group != channel_group_to_download:
                continue
```

Examples/download_studies.py
```diff
@@ -0,0 +1,79 @@
...
+def get_download_status_file(client):
+    """Retrieves JSON of download status for all study IDs."""
+    studies_file = join(dirname(__file__), 'studies.json')
+    if not isfile(studies_file):
+        status = {
+            study_id: 0
+            for study_id in client.get_study_ids()
+        }
+        write_json(studies_file, status)
+    return studies_file
```
bjdoyle, line 23: It would be better to get this function to return the dict, then you wouldn't need to write the empty file, just return an empty dict if the file doesn't exist, or return the json from the file if it does.

Examples/download_studies.py
```diff
@@ -0,0 +1,79 @@
...
+            return
+        output_dir = join(path_out, 'data')
+    else:
+        output_dir = join(dirname(__file__), 'data')
+    if not isdir(output_dir):
+        mkdir(output_dir)
+
+    download_status_file = get_download_status_file(client)
+    download_status = read_json(download_status_file)
+    study_ids = [study_id for study_id in download_status if download_status[study_id] == 0]
```
bjdoyle, line 40:
```suggestion
    study_ids = [study_id for study_id in download_status if not download_status[study_id]]
```

Examples/download_studies.py
```diff
@@ -0,0 +1,79 @@
...
+            for study_id in study_ids:
+                downloader = DataDownloader(client, study_id, output_dir)
+                # Download channel data
+                downloader.download_channel_data(channel_group_to_download)
+                download_status[study_id] = 1
+                write_json(download_status_file, download_status)
+            print('Download completed.')
+            return
+
+        except ReadTimeoutError:
```
bjdoyle, line 57: If you are downloading multiple studies this will re-download previous studies if there is an error on any study after the first one. You either need to have this except statement inside the for study_id in study_ids loop, or you need to do the check for whether a study has been downloaded already in the for loop.

bjdoyle (changes requested):

Examples/download_studies.py
dkeden: Hey @bjdoyle, could you please let me know what you think of the most recent change? Would I need to move the `if attempts == 5` line to inside the `except` statement?

dkeden:
Hey @norrishd, any chance you could review this? Brendan is away until tomorrow, and am hoping to have this script available today! No worries if not, let me know

Examples/download_studies.py
```diff
@@ -0,0 +1,78 @@
...
+
+def run(client, path_out, channel_group_to_download):
+    """Downloads all data available on Seer's public API."""
+    if path_out:
+        if not isdir(path_out):
+            print('Please specify a path that exists.')
+            return
+        output_dir = join(path_out, 'data')
+    else:
+        output_dir = join(dirname(__file__), 'data')
```
norrishd, line 34: I wonder if it would be more intuitive to a new user for the data to default to downloading to `~/data` i.e. in their current directory rather than putting it in `seer-py/Examples/data`?

Examples/download_studies.py
```diff
@@ -0,0 +1,78 @@
...
+from os.path import dirname, isdir, isfile, join
+from urllib3.exceptions import ReadTimeoutError
+from downloader.utils import read_json, write_json
+from seerpy import SeerConnect
+from downloader.downloader import DataDownloader
+
+
+def get_download_status_file(client):
+    """Retrieves JSON of download status for all study IDs."""
+    studies_file = join(dirname(__file__), 'studies.json')
```
norrishd, line 16: Similar to a comment below, it might be better not to download data into a version-controlled directory (the ol' "separate your data from your source files" maxim). The usual approach I've encountered with other packages is they create a public or private folder in the user's home directory and store stateful stuff in there, e.g. maybe `~/seerpy`.
https://stackoverflow.com/a/4028943/8822495

Examples/download_studies.py
```diff
@@ -0,0 +1,78 @@
...
+
+from os import mkdir
+from os.path import dirname, isdir, isfile, join
+from urllib3.exceptions import ReadTimeoutError
+from downloader.utils import read_json, write_json
+from seerpy import SeerConnect
+from downloader.downloader import DataDownloader
+
+
+def get_download_status_file(client):
```
norrishd, line 14: One risk here is that study IDs get downloaded once then never again, so if/when new studies get created, the user won't be able to download them and might not understand why.

If this is just intended as a one-off script that's probably fine, otherwise it might be worth adding another `argparse` flag to force re-download of study IDs, which can get passed to this function and handled.

Examples/download_studies.py
```diff
@@ -0,0 +1,78 @@
...
+    download_status = read_json(download_status_file)
+    study_ids = [study_id for study_id in download_status if not download_status[study_id]]
+
+    for study_id in study_ids:
+        attempts = 0
+        while True:
+            try:
+                if attempts != 0:
+                    time.sleep(3)
+                    client = SeerConnect()
```
norrishd, line 48: This is more for my own edification, but what's the advantage in re-instantiating a `SeerConnect()` with each timeout?

Examples/download_studies.py
```diff
@@ -0,0 +1,78 @@
...
+                downloader = DataDownloader(client, study_id, output_dir)
+                # Download channel data
+                downloader.download_channel_data(channel_group_to_download)
+                download_status[study_id] = 1
```
norrishd, line 52: If this script is invoked once with `--channelgroup eeg` (or whatever is valid) then re-run with a different channel group, won't download anything.

What if the `download_status` file instead stored e.g.
```
status = {
    study_id1: ['eeg', 'ecg', ... ],
    study_id2: ['eeg', 'ecg', ... ]
}
```
Again maybe this is just meant to be a quick and dirty script, in which case all good!

Examples/download_studies.py
```diff
@@ -0,0 +1,78 @@
...
+                write_json(download_status_file, download_status)
+
+            except ReadTimeoutError:
+                attempts += 1
+                for i in range(5, 0, -1):
+                    sys.stdout.write(f'ReadTimeoutError. Re-trying after a short break... {str(i)}')
+                    sys.stdout.flush()
+                    time.sleep(1)
```
norrishd, line 60: I've never come across `sys.stdout.write/flush`, but for me when I execute these lines in a script I get all on one line:
```ReadTimeoutError. Re-trying after a short break... 5ReadTimeoutError. Re-trying after a short break... 4ReadTimeoutError. Re-trying after a short break... 3ReadTimeoutError. Re-trying after a short break...```

Is that the intention?

Comment (norrishd) [reply Examples/download_studies.py:60]: Oh I get it!
https://stackoverflow.com/a/10019596/8822495

That's so interesting, I remember noticing on Ubuntu that if you do:
```python
for i in range(5):
    print(i, end='')
    time.sleep(1)
```
It sits there for 5 seconds then prints it all at once. I always thought this was a bug with the `end` param. But I guess you're addressing this kind of issue that crops up on some systems?

What if the "ReadTimeoutError. Re-trying after a short break..." was printed on its own line before the loop, then in the loop just `str(i)` is printed?

Examples/download_studies.py
```diff
@@ -0,0 +1,78 @@
...
+            print('Download completed.')
+            return
+
+        # If user reaches 5 attempts, break and request user to return later
+        if attempts == 5:
```
norrishd, line 65: Out of curiosity how common are timeouts? Does it ever timeout 2 or 3 times then make it on the 4th/5th try?

norrishd (approved):
Made a bunch of comments but maybe they're all overkill.

PR #123, opened 2021-01-13
Move matplotlib and scipy into optional dependencies

norrishd:
Remove `scipy` from the requirements, along with the 4 filter-related helper function in _utils.py_ that use it. Also change `matplotlib` to be an optional dependency. These two changes make `seer-py` small enough that it can be used as-is in AWS lambda functions (e.g. as a layer or as part of a code + dependencies zipfile).

To enable the `plot_eeg()` function, install with:
`pip install .[viz]`


README.md
```diff
@@ -7,7 +7,7 @@ Note that by default there is no public access to any data on the Seer Cloud. Yo
 
 ## Install
 
-To install, simply clone or download this repository, then type `pip install .` which will install all the dependencies.
+To install, simply clone or download this repository, then type `pip install .` which will install all the dependencies. To enabling plotting signal data (required by the example notebook), use `pip install .[viz]`
```
bjdoyle, line 10:
```suggestion
To install, simply clone or download this repository, then type `pip install .` which will install all the dependencies. To enable plotting signal data (required by the example notebook), use `pip install .[viz]`
```

bjdoyle (dismissed):

bjdoyle (approved):
I hereby enabling your merge.

PR #125, opened 2021-01-20
Update json_normalize method to remove warning

dkeden:
Are we able to update the method `json_normalize` on Pandas to avoid the warning @bjdoyle? I don't think we have a specified version of Pandas in requirements.


bjdoyle (changes requested):
This only moved in version 1.0.0, I wouldn't change it yet.
When we do we would need to add >=1.0.0 to the pandas requirement.

norrishd:
Resolved by #141

PR #126, opened 2021-02-15
Cope with missing data and s3 performance errors when getting data.

bjdoyle:
2 related changes:
1. change seerpy.get_data_chunk_urls() to cope with missing data chunks (resulting from processing errors)
2. change utils.download_channel_data() to cope with missing data chunks (these are the result of upload errors) and also with s3 performance errors (it throws http 500 and 503 errors when it gets overloaded). In the case of 500 and 503 we will retry up to twice.
Also fixed a couple of errors in docstrings.

### DIFF
--- a/seerpy/utils.py
+++ b/seerpy/utils.py
@@ -65,12 +65,12 @@ def download_channel_data(data_q, download_function):
 
         if meta_data['channelGroups.timestamped']:
             # timestamped data is not stored in records as EDF data is
-            # it is just a sequence of (ts, ch1, ch2, ..., chN), (ts, ch1, ch2, ..., chN), ...
+            # it is just a sequence of (ts1, ch1, ch2, ..., chN), (ts2, ch1, ch2, ..., chN), ...
             # the timestamp is milliseconds relative to chunk start
             column_names = ['time'] + channel_names
             data = data.reshape(-1, len(column_names))
         else:
-            # EDF data is in the format [record 1: (ch1 sample1, ch1 sample2, ..., ch2 sampleN),
+            # EDF data is in the format [record 1: (ch1 sample1, ch1 sample2, ..., ch1 sampleN),
             # (ch2 sample1, ch2 sample2, ..., ch2 sampleN), ...][record2: ...], ..., [recordN: ...]
             data = data.reshape(-1, len(channel_names),
                                 int(meta_data['channelGroups.samplesPerRecord']))
@@ -293,7 +293,7 @@ def get_channel_data(study_metadata, segment_urls, download_function=requests.ge
                                 na_position='last')
         data = data.reset_index(drop=True)
     else:
-        data = None
+        data = pd.DataFrame()
 
     return data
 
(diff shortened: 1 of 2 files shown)


seerpy/utils.py
```diff
@@ -140,18 +134,82 @@ def download_channel_data(data_q, download_function):
...
+            print('The chunk requested does not exist')
+            return None
+
+        if status_code in (500, 503):
+            # s3 sometimes returns 500 or 503 if it's overwhelmed, we can retry
+            message = 'Unable to read chunk - most likely a performance error'
+            if i < (max_attempts - 1):
+                print(message, '- retrying')
+                continue
+            print(message, '- max attempts exceeded')
```
norrishd, line 199: Probably good to sleep for a bit with backoff in the case of an exception? (especially for 503, which is returned for [SlowDown and ServiceUnavailable](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html) error codes)

seerpy/utils.py
```diff
@@ -140,18 +134,82 @@ def download_channel_data(data_q, download_function):
...
+
+        # throw an error for other status codes
+        raise requests.exceptions.HTTPError(f'HTTPError {status_code}')
```
norrishd, line 202: Since AWS returns the same few status codes for multiple problems, might be helpful for debugging to also print the `response.reason`.

This might be overkill here, but for debugging Asset Panda requests I also found there was useful information in the body (though sometimes the body is empty), so I added

```python
try:
    body = response.json()
except json.JSONDecodeError:
    body = {}
# Add the body to the error message logged
```

norrishd (dismissed):
Nice PR 👍 Added a couple of small suggestions which you can take or leave.

Other random thought: have we ever tried multi-threaded downloading from S3/CloudFront? S3 supports >= 5,500 GET/HEAD requests per second per prefix in an S3 bucket, and it's super easy to implement with `concurrent.futures`. Though I don't know what the prefix structure is in S3, so I guess this could be risky if there are someday lots of users downloading from the same prefix

seerpy/utils.py
bjdoyle: That would be more robust, but at the moment I've only seen individual 500 errors (which I think are an accidental form of 503) and they've always recovered after a single retry with no sleep, so I'll keep it simple for the moment but monitor how it goes.

seerpy/utils.py
bjdoyle: Good idea I might add that.

bjdoyle: Actually the AWS responses are XML so I'm just logging the content as is.

norrishd: TIL!

norrishd (approved):
Cha-ching! 💰

seerpy/utils.py
```diff
@@ -6,6 +6,7 @@
 """
 import functools
 import gzip
+import json
```
norrishd, line 9: Oh one thing, this isn't used anymore

bjdoyle: Motherf@#$%er!

bjdoyle: OK, take 3!

bjdoyle: Previous comment aimed at the state of the universe, not you, in case there was any doubt.

norrishd (approved):
Nice, sorry I didn't notice this til after approving last time

PR #128, opened 2021-03-04
Update setup instructions to make them clearer

cfranklin11:
I found the setup/installation instructions a bit difficult to follow, so I put them toward the top and updated them to reflect the current state of the package's dependencies. I also grouped the data info together.


EwanNurse (approved):
The restructure of this readme makes sense and makes the document more readable.

PR #129, opened 2021-03-05
Drop samples that exceed number of samples per record

cfranklin11:
Some EDF files have a number of samples that don't divide evenly
into the expected number of samples per record, which raises
errors when we try to reshape the data. The underlying cause is
unclear, but since the tendency is for there to be fewer than
200 extra samples, it's easier to drop them rather than pad the
data with blank values. We still print a warning in case the user
wants to go back and work with the offending file directly.

### DIFF
--- a/seerpy/utils.py
+++ b/seerpy/utils.py
@@ -8,6 +8,7 @@ import functools
 import gzip
 from multiprocessing import Pool
 import os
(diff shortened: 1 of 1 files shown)


cfranklin11:
This got properly fixed in #144

PR #132, opened 2021-03-22
Resolves #130 : Limit max items returned in paginated query

### LINKED ISSUE #130: Feature Request: Limit max items returned in paginated query
https://github.com/seermedical/seer-py/blob/ec2599f0392f277c5bbf1b2643104e0b9dc06725/seerpy/seerpy.py#L158

Can I suggest that we add a parameter to the `client.get_paginated_response()` that limits the total amount of items that it returns? Currently we have `limit` which limits the amount of items returned per page in the paginated request. But nothing to limit the total number of items returned. 

I am happy to implement this. I just want to get peoples input on this idea, and what the argument should be called. Can I suggest calling it `max_items`?

**NOTE:** for queries such as the following, which contain nested lists of items, the `max_items` would only affect the outer list (2 `users`, irrespective of how many `surveys` for those users).

```
query userSurveysLimit{
  users(offset: 82, limit:2){
    username
    surveys{
        id
    }
  }
}
```

**Where it is useful**

When getting studies, we currently have `client.get_studies()` which gets ALL studies that the user has access to. If a user or developer is just experimenting, they may not want to wait around for 10 or more minutes to get ALL studies. Its an unpleasant experience.

**Upstream changes**

If this idea is accepted, I suggest also adding the same arguments to all other functions that call `get_paginated_response`, such as `get_studies()` `get_studies_dataframe()`, etc.

I am happy to implement this as well.

Besides the functions in `seerpy.py` are there any other calls to `client.get_paginated_response()` that i should know about?

ronrest:
I have added the `max_items` parameter to the following functions.

```python
get_paginated_response()

get_studies()
get_studies_dataframe()
get_patients()
get_mood_survey_results()
get_mood_survey_results_dataframe()
get_labels()
get_labels_dataframe()
get_diary_insights()
get_user_ids_in_user_cohort()
get_study_ids_in_study_cohort()
get_diary_study_labels()
get_diary_study_labels_dataframe()
```

## Ommited

I have not added `max_items` to the following functions

```python
get_organizations()                 # number of organizations is too few, so didnt implement
                                    # pagination
get_organisations_dataframe()

get_all_bookings()                  # Underlying GraphQL query for Bookings does not support
                                    # pagination. But i did allow  for a `includeCancelled` option
                                    # to be used
get_all_bookings_dataframe()        # As above

get_diary_study_label_groups()      # As per a TODO comment in the function, its unlikely that we
                                    # will get more than 20 items
get_diary_study_label_groups_dataframe() # As above

get_diary_labels()                  # Contains nesting that would make it unintuitive to use
                                    # `max_items`
                                    # eg, for each user, the results can contain contain multiple
                                    # `labelGroups` and for each `labelGroup` there can be multiple
                                    # `labels`
get_diary_labels_dataframe()        # As above


get_diary_medication_alerts()       # Contains nesting that would make it unintuitive to use
                                    # `max_items`
                                    # multiple items at the ["alerts"] level
                                    # multiple items at the ["alerts", "labels"] level
                                    # multiple items at the ["alerts", "labels", "doses"] level
get_diary_medication_alert_windows()# Contains nesting that would make it unintuitive to use
                                    # `max_items`
                                    # multiple items at ["alerts"] level
                                    # multiple items at ["alerts", "windows"] level
```

Also, ones that we could potentially add the option to in the future

```python
get_diary_study_channel_groups()    # currenlty it runs as a single query call.
                                    # how many channel groups are there for diary studies?
                                    # could it benefit from having a max_limit

get_diary_study_channel_groups_dataframe() # As above
```

## Testing

I have performed some preliminary sanity checks on the following:

```python
get_studies()
get_studies_dataframe()
get_patients()
get_mood_survey_results()
get_mood_survey_results_dataframe()
```

Ones I havent tested so far

```python
get_labels()
get_labels_dataframe()

get_diary_insights()                # Havent found any records with diary insights to check
get_user_ids_in_user_cohort()       # Havent found any any cohort ids to check
get_study_ids_in_study_cohort()     # havent found any study cohort ids to check

get_diary_study_labels()            # havnt found any patients with diary study labels to check
get_diary_study_labels_dataframe()  # as above
```

### DIFF
--- a/seerpy/graphql.py
+++ b/seerpy/graphql.py
@@ -408,6 +408,27 @@ GET_PATIENTS = """
         }
     }"""
 
+
+GET_PATIENTS_PAGED = """
+    query patientsPaged(
+                  $limit: PaginationAmount,
+                  $offset: Int) {
+        patients (limit: $limit, offset: $offset) {
+          id
+          user {
+              id
+              fullName
+              shortName
+              email
+              lastActive
+              lastInsightGenerated
+              preferredTimezone
+          }
+        }
+    }
+    """
+
+
 GET_DIARY_INSIGHTS_PAGED = """
     query patient($patient_id: String!,
                   $limit: PaginationAmount,
--- a/seerpy/graphql.py
+++ b/seerpy/graphql.py
@@ -671,6 +671,55 @@ def get_bookings_query_string(organisation_id, start_time, end_time):
             }""" % (organisation_id, start_time, end_time)
 
 
+# NOTE: This provides more flexibility than using `get_bookings_query_string()`
+#       eg, allowing null values for variables.
+# But it is not plug and play compatible with the returned value of that
+# function, so keeping this separate.
+# TODO: maybe remove get_bookings_query_string()
+ORGANIZATION_BOOKINGS = """
+    query organizationBookings($organization_id: String!, $startTime: Float, $endTime: Float, $includeCancelled: Boolean = false){
+        organisation(id: $organization_id) {
+            bookings(startTime: $startTime, endTime: $endTime, includeCancelled: $includeCancelled) {
+                id
+                equipmentItems {
+                    name
+                    equipmentType {
+                        type
+                    }
+                }
+                bookingTemplate {
+                    name
+                }
+                referral {
+                    id
+                }
+                startTime {
+                    datetime
+                    timezone
+                }
+                endTime {
+                    datetime
+                    timezone
+                }
+                patient {
+                    id
+                    user {
+                        fullName
+                    }
+                    studies {
+                        id
+                        name
+                    }
+                }
+                location {
+                        name
+                        suburb
+                        }
+            }
+        }
+    }"""
+
+
 def get_diary_study_label_groups_string(patient_id, limit, offset):
 
     return """
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -577,7 +577,8 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
         study_df: pd.DataFrame
             DataFrame with details of all matching studies
         """
-        studies = self.get_studies(limit=limit, search_term=search_term, party_id=party_id, max_items=max_items)
+        studies = self.get_studies(limit=limit, search_term=search_term, party_id=party_id,
+            max_items=max_items)
         studies_dataframe = json_normalize(studies).sort_index(axis=1)
         return studies_dataframe.drop('patient', errors='ignore', axis='columns')
 
@@ -787,7 +788,8 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
         return data_chunk_urls
 
     # pylint:disable=too-many-arguments
-    def get_labels(self, study_id, label_group_id, from_time=0, to_time=9e12, limit=200, offset=0, max_items=None):
+    def get_labels(self, study_id, label_group_id, from_time=0, to_time=9e12, limit=200, offset=0,
+        max_items=None):
         """
         Get labels for a given study and label group.
 
@@ -1625,7 +1627,8 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
 
         return response['organisation']['bookings']
 
-    def get_all_bookings_dataframe(self, organisation_id, start_time, end_time, include_cancelled=False):
+    def get_all_bookings_dataframe(self, organisation_id, start_time, end_time,
+        include_cancelled=False):
         """
         Get all bookings for any studies that are active at any point between
         `start_time` and `end_time` as a DataFrame. See `get_all_bookings()`.
@@ -1915,7 +1918,8 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
                                            max_items=max_items
                                            )
 
-    def get_mood_survey_results_dataframe(self, survey_template_ids, limit=200, offset=0, max_items=None):
+    def get_mood_survey_results_dataframe(self, survey_template_ids, limit=200, offset=0,
+        max_items=None):
         """
         Get mood survey results as a DataFrame. See `get_mood_survey_results()`
         for details.
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -234,14 +234,14 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
                                           party_id=party_id)
 
             # select the part of the response we are interested in
-            response = utils.get_nested_dict_item(response, *object_path)
+            response = utils.get_nested_dict_item(response, object_path)
 
 
             # select the part of the response which can vary. if iteration_path is None this will be
             # the same as the part of the response we are interested in
             response_increment = response
             if iteration_path:
-                response_increment = utils.get_nested_dict_item(response_increment, *iteration_path)
+                response_increment = utils.get_nested_dict_item(response_increment, iteration_path)
             if not response_increment:
                 # if the part of the response which varies is empty, we are finished iterating
                 break
@@ -257,7 +257,7 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
                 # otherwise add the response increment to the existing result at the correct level
                 values_container = result
                 if iteration_path:
-                    values_container = utils.get_nested_dict_item(values_container, *iteration_path)
+                    values_container = utils.get_nested_dict_item(values_container, iteration_path)
                 values_container.extend(response_increment)
 
             offset += limit
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -1616,6 +1616,7 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
             'patient', 'referral', 'equipmentItems', and 'location'
         """
         # TODO: paginate using the new resource schema
+        assert (start_time is not None) and (end_time is not None), "start_time and end_time should not be None"
         query_string = graphql.ORGANIZATION_BOOKINGS
         vars = dict(organization_id=organisation_id, startTime=start_time, endTime=end_time,
                     includeCancelled=include_cancelled)
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -184,12 +184,12 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
             vary with each query iteration. If None then the response varies at
             the path given by object_path.
 
-            Eg: We may be interested in the surveys for a user. The heirarchy
-            will look something like ["user", "surveys"]. We are interested in
-            keeping the user information, although this doesnt change with each
-            iteration, so we set `object_path=["user"]`. But we actually want
-            to iterate through the survey  information, which does change with
-            each iteration, so we set `iteration_path=["surveys"]`
+            e.g. We may be interested in the surveys for a user, while keeping
+            information about the user. The hierarchy will look something like
+            ["user", "surveys"]. At the "user" level, it contains the user
+            information which we want to keep, so we set `object_path=["user"]`.
+            But we cannot iterate at this level, we want to iterate at the
+            "surveys" level, so we set `iteration_path=["surveys"]`
         party_id : str, optional
             The organisation/entity to specify for the query
         max_items: int, optional
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -1615,7 +1615,7 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
             Booking information, with keys including 'id', 'startTime', 'endTime',
             'patient', 'referral', 'equipmentItems', and 'location'
         """
-        # TODO: request changes to seer-api so we can perform pagination
+        # TODO: paginate using the new resource schema
         query_string = graphql.ORGANIZATION_BOOKINGS
         vars = dict(organization_id=organisation_id, startTime=start_time, endTime=end_time,
                     includeCancelled=include_cancelled)
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -171,7 +171,8 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
         variable_values : dict, optional
             Values for GraphQL to substitute into the query
         limit : int
-            Batch size for repeated API calls
+            Batch size for repeated API calls. Does not affect the total number
+            of items retrieved
         object_path : list of str
             One or more levels of key giving the path to the object to be returned
             e.g. ['userCohort', 'users'] for a query response of
@@ -801,7 +802,8 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
         to_time : int, optional
             Timestamp in msec - only retrieve data up until this point
         limit : int, optional
-            Batch size for repeated API calls
+            Batch size for repeated API calls. Does not affect the total number
+            of items retrieved
         offset : int, optional
             Index of first label to retrieve
         max_items: int, optional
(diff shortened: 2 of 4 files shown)


seerpy/graphql.py
```diff
@@ -650,6 +671,55 @@ def get_bookings_query_string(organisation_id, start_time, end_time):
             }""" % (organisation_id, start_time, end_time)
 
 
+# NOTE: This provides more flexibility than using `get_bookings_query_string()`
+#       eg, allowing null values for variables.
+# But it is not plug and play compatible with the returned value of that
+# function, so keeping this separate.
+# TODO: maybe remove get_bookings_query_string()
+ORGANIZATION_BOOKINGS = """
```
norrishd, line 679: IMO this is a good improvement over `get_bookings_query_string()`. I have a couple of uncertainties:

1. Maybe it's a bit fraught to allow null values for `startTime` and `endTime`? There are thousands of bookings, and with no pagination as well...
1. Did you tune in for the workshop on the new "resources" schema? It allows for pagination for bookings, and if we're going to update queries in `seer-py` I think @cheewee-seer would approve of us switching to the new schema. That might be a slightly bigger piece though...

seerpy/seerpy.py
```diff
@@ -170,57 +171,94 @@ def get_paginated_response(self, query_string, variable_values, limit, object_pa
...
-            path given by object_path
+            None (default), one, or more levels of key giving the path to the
+            node (relative to the `object_path` level) where the response can
+            vary with each query iteration. If None then the response varies at
+            the path given by object_path.
+
+            Eg: We may be interested in the surveys for a user. The heirarchy
+            will look something like ["user", "surveys"]. We are interested in
+            keeping the user information, although this doesnt change with each
+            iteration, so we set `object_path=["user"]`. But we actually want
```
norrishd, line 190: Nits: "e.g." is used just above, so would be good to be consistent throughout the documentation.

The example is really helpful, though I'm a tiny bit confused by the comment - is it giving two different examples? "But we actually" reads to me like the first prescription is actually wrong. Maybe a slight rephrasing to "If we are interested in... If we instead want to iterate through..."

Also there's a double space with "survey  information" :-)

seerpy/seerpy.py
```diff
@@ -170,57 +171,94 @@ def get_paginated_response(self, query_string, variable_values, limit, object_pa
...
+            to iterate through the survey  information, which does change with
+            each iteration, so we set `iteration_path=["surveys"]`
         party_id : str, optional
             The organisation/entity to specify for the query
+        max_items: int, optional
+            max number of items to return. In the case of queries containing
```
norrishd, line 196: More nits: Probably good to capitalise the first word ("Max") to be consistent with the documentation of other args. This goes for all the other methods where `max_items` has been added.

seerpy/seerpy.py
```diff
@@ -170,57 +171,94 @@ def get_paginated_response(self, query_string, variable_values, limit, object_pa
...
+            querying a single user, and the list of surveys for this user, then
+            it limits the number surveys to 10 for this user, while still
+            returning the information at the `user` level.
 
         Returns
         -------
         responses: list of dict
             List of query result dictionaries
         """
+        variable_values = deepcopy(variable_values) # prevent local changes affecting external one
```
norrishd, line 215: Nits: PEP likes two spaces before in-line comments. Ditto the comment just below

seerpy/seerpy.py
```diff
@@ -170,57 +171,94 @@ def get_paginated_response(self, query_string, variable_values, limit, object_pa
...
-                for key in iteration_path:
-                    response_increment = response_increment[key]
+                response_increment = utils.get_nested_dict_item(response_increment, *iteration_path)
             if not response_increment:
                 # if the part of the response which varies is empty, we are finished iterating
                 break
 
+            # Update the number of items received
+            total_items_returned += len(response)
+            # print(f"items: {len(response)}  limit: {limit}  fetched: {total_items_returned} remain: {remaining_items}") # for debuging purposes
```
norrishd, line 251: Accidentally left this in!

seerpy/seerpy.py
```diff
@@ -1009,38 +1069,55 @@ def get_user_from_patient_dataframe(self, patient_id):
             return pd.DataFrame()
         return json_normalize(patient).sort_index(axis=1)
 
-    def get_patients(self, party_id=None):
+    def get_patients(self, party_id=None, limit=50, max_items=None):
```
norrishd, line 1072: Did you do any experimenting with whether 50 is a good size? My intuition is that we could probably go a fair bit higher to improve performance, though I'm not sure who to check this with.

seerpy/seerpy.py
```diff
@@ -1521,18 +1608,27 @@ def get_all_bookings(self, organisation_id, start_time, end_time):
...
+            Whether to include cancelled bookings
         Returns
         -------
         bookings : list of dict
             Booking information, with keys including 'id', 'startTime', 'endTime',
             'patient', 'referral', 'equipmentItems', and 'location'
         """
-        query_string = graphql.get_bookings_query_string(organisation_id, start_time, end_time)
-        response = self.execute_query(query_string)
+        # TODO: request changes to seer-api so we can perform pagination
```
norrishd, line 1619: See my other comment - we can perform pagination :-)

seerpy/seerpy.py
```diff
@@ -1521,18 +1608,27 @@ def get_all_bookings(self, organisation_id, start_time, end_time):
...
+        query_string = graphql.ORGANIZATION_BOOKINGS
+        vars = dict(organization_id=organisation_id, startTime=start_time, endTime=end_time,
+                    includeCancelled=include_cancelled)
+        response = self.execute_query(query_string, variable_values=vars)
+
+        # query_string = graphql.get_bookings_query_string(organisation_id, start_time, end_time)
+        # response = self.execute_query(query_string)
```
norrishd, line 1626: Good not to leave commented-out code in there

seerpy/seerpy.py
```diff
@@ -1508,7 +1595,7 @@ def get_channel_data(self, all_data, segment_urls=None, download_function=reques
         return utils.get_channel_data(all_data, segment_urls, download_function, threads, from_time,
                                       to_time)
 
-    def get_all_bookings(self, organisation_id, start_time, end_time):
+    def get_all_bookings(self, organisation_id, start_time, end_time, include_cancelled=False):
```
norrishd, line 1598: Do you know if the previous default would have implicitly been `True` for including cancelled bookings? If so this means a breaking change for consumers, so we'd need to bump the major version number. If it's not breaking we should bump the minor version number as new functionality has been added (in _setup.py_) - https://semver.org/

seerpy/seerpy.py
```diff
@@ -1521,18 +1608,27 @@ def get_all_bookings(self, organisation_id, start_time, end_time):
...
         Returns
         -------
         bookings : list of dict
             Booking information, with keys including 'id', 'startTime', 'endTime',
             'patient', 'referral', 'equipmentItems', and 'location'
         """
-        query_string = graphql.get_bookings_query_string(organisation_id, start_time, end_time)
-        response = self.execute_query(query_string)
+        # TODO: request changes to seer-api so we can perform pagination
+        query_string = graphql.ORGANIZATION_BOOKINGS
```
norrishd, line 1620: Subjective opinion: could just pass this into the `execute_query()` call without first assigning to a variable? Your call though, I don't really mind.

seerpy/utils.py
```diff
@@ -522,3 +522,99 @@ def quote_str(value):
...
+    func:   function object
+        The function to run
+    *args:
+        positional arguments to pass to `func`
+    **kwargs:
+        keyword arguments to pass to `func`
+
+    Returns
+    ----------
+    bool
```
norrishd, line 546: What is the purpose of adding this function? I can't see anywhere else in the repo that it's used.

seerpy/utils.py
```diff
@@ -522,3 +522,99 @@ def quote_str(value):
...
+
+def get_nested_dict_item(d, *keys, allow_missing_keys=False, default=None):
+    """
+    Given a dictionary with potentially many nested dictionaries, get the
+    value of one of the nested keys by providing the sequence of keys as
+    arguments.
+
+    Parameters
+    ----------
+    d: dict
```
norrishd, line 576: A more descriptive variable name than `d` would be good. I know it's extremely obvious what it is, and there's an immediate description, but it's a general good practice.

seerpy/utils.py
```diff
@@ -522,3 +522,99 @@ def quote_str(value):
...
+        The dictionary
+    *keys:
```
norrishd, line 578: Would it be safer/clearer for this to take a list of str? IDK if there's any risk of `allow_missing_keys` and `default` being mixed in with the other keys - there might not be, in which case feel free to delete this comment!

seerpy/utils.py
```diff
@@ -522,3 +522,99 @@ def quote_str(value):
...
+        The sequence of keys to traverse along the heirarchy of the dictionary
+    allow_missing_keys: bool, optional
+        Allow travering along missing keys? If so, this acts like the
```
norrishd, line 581: "traversing"

norrishd (changes requested):
Awesome work and thanks for the proactive approach in improving the repo!

I've made a number of changes, most of them small nit-picky things.

However more importantly, a number of the tests failed on Buildkite - https://buildkite.com/seer-medical/seer-py/builds/658#9195c5c2-7e8e-4729-839c-8d8cb932753f

I'm not sure if they're due to your changes to `get_paginated_response()` or were previously failing and no one noticed, but would you be able to double check?

seerpy/utils.py
ronrest: I dont think there is a danger of that, `allow_missing_keys` and `default` need to be passed in as keyword arguments. But, as you say,  it is perhaps cleaner to have the keys as a list of strings. It will allow a user to pass a variable containing the list of keys without having to prefix it with a `*` to unpack the values.

ronrest: fixed this

seerpy/utils.py
ronrest: fixed this

seerpy/utils.py
ronrest: This is a tricky one. 

- I cant have it as `dict` because that is a reserved name. 
- I changed it to `dct`, which im not sure is really any better than just `d`
- I could have something like `input_dict`, but then it starts to get a bit too verbose

Do you have any other suggestions for the name?

seerpy/utils.py
ronrest: Was essentially there for testing that a function throws the expected Error. But unnecessary now that you have pointed out the `pytest` library.

Deleted this function

seerpy/seerpy.py
ronrest: I think it keeps it consistent with all the other functions to have this.

seerpy/seerpy.py
ronrest: fixed

seerpy/seerpy.py
ronrest: Agreed, it was a little confusing. I rewrote it as follows, do you think it makes more sense now? 

```
e.g. We may be interested in the surveys for a user, while keeping
information about the user. The hierarchy will look something like
["user", "surveys"]. At the "user" level, it contains the user
information which we want to keep, so we set `object_path=["user"]`.
But we cannot iterate at this level, we want to iterate at the
"surveys" level, so we set `iteration_path=["surveys"]`
```

seerpy/seerpy.py
ronrest: fixed

seerpy/seerpy.py
ronrest: fixed

seerpy/seerpy.py
ronrest: fixed

seerpy/graphql.py
ronrest: 1. added an assert statement as a temporary measure to protect against this while it remains non-paginated.
  ```assert (start_time is not None) and (end_time is not None), "start_time and end_time should not be None"```
2. This should maybe be done in a separate pull request. It will be a bigger task.

seerpy/seerpy.py
ronrest: We should treat the migration to resource schema as a separate pull request

seerpy/seerpy.py
ronrest: The default in the graphql api was already `False`

image

seerpy/seerpy.py
norrishd: Yep I get it now!

seerpy/seerpy.py
norrishd: Yep agreed! Matias is interested in this too.

seerpy/utils.py
norrishd: Haha yeah variable naming, one of the two hard problems! Anything like `input_dict`, `nested_dict`, `dict_of_dict` seems fine to me.

seerpy/graphql.py
norrishd: Agreed, that should be a separate PR 👍

seerpy/seerpy.py
ronrest: You are correct. Setting it at a limit of `50` is painfullly slow. Running this at

Running this a few times at different `limit` values and averaging the time it takes to get ALL patients we get:

- At a value of `20,00` or above, it times out.
- At a value of `15,000` it sometimes times out.
- At a value of `10,000` it is the quickest, but could potentially time out for each page if the network is slow.
- At a value of `5,000`, it is only a bit slower than setting it to `10,000`, but potentially much safer.

| limit | time_secs | notes |
|------:|----------:| :--- |
| 15000 |   91.966 | high due to some timeouts |
| 10000 |   31.935 |  |
|  8000 |   33.132 |  |
|  7000 |   36.382 |  |
|  6000 |   38.010 |  |
|  5000 |   38.728 |  |
|  4000 |   41.985 |  |
|  3000 |   46.052 |  |
|  2000 |   46.223 |  |
|  1000 |   54.295 |  |
|   500 |  126.446 |  |
|   200 |  232.004 |  |
|   100 |  347.360 |  |
|    50 |  705.313 |  |


So setting limit to `5000`

seerpy/utils.py
ronrest: changed it to `input_dict `

ronrest:
Ready for re-review. 

It passes all the tests locally. Not sure how to trigger the build-kite tests though.

seerpy/seerpy.py
```diff
@@ -483,19 +520,22 @@ def get_study_ids(self, limit=50, search_term='', party_id=None):
...
-            Batch size for repeated API calls
+            Batch size for repeated API calls. Does not affect the total number
+            of items retrieved
         search_term : str
             Filter results to studies including this string, either in the study
             name or patient name. Not case sensitive.
         party_id : str
             The organisation/entity to specify for the query
+        max_items: int, optional
+            max number of studies to return.
```
norrishd, line 538: No worries if you don't change, but would be good to capitalise the "max" in the docstrings for all the functions that now have `max_items`

norrishd (approved):
Looks great! Could you please also bump the minor version number in _setup.py_?

seerpy/seerpy.py
ronrest: Ooops, i missed that. Fixed now :)

ronrest:
Bumped up the version number to 0.5.0

norrishd (approved):
Woop woop!

matias-seer (approved):
Looks good. thanks for making the effort to improves these functions

PR #133, opened 2021-03-29
Avoid using empty SDK API URL from env vars

norrishd:
# Description
If an empty `SDK_API_BASE_URL` environment variable is found, `seer-py` will try using that as the auth `api_url` and fail. This has caused a little trouble in testing lambdas using AWS SAM, as they export this variable as empty when the `Env` parameter isn't set - which it isn't for local testing.

This tiny change will fall back to the `f'https://sdk-{region}.seermedical.com/api'` endpoint in that case.

For example, see:
https://github.com/seermedical/seer-api-events/blob/8e74354ebc22d28d2555d8726b93392423d774e3/apps/tedsZendeskTickets/template.yaml#L72

### DIFF
--- a/seerpy/auth.py
+++ b/seerpy/auth.py
@@ -303,7 +303,7 @@ class SeerApiKeyAuth(BaseAuth):
                                                                     region)
 
         if not api_url:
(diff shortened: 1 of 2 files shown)


seerwill (approved):
LGTM

PR #134, opened 2021-04-04
Added a new graphql call to fetch >5000 segment IDs 

matias-seer:
A new function to fetch segment ids for a channel group contining more than 5000 segments. The new function 'get_segment_ids' return a dataframe containing all segment ids in the channel group.

note: an existing function, 'get_paginated_response' could be used instead to achieve a similar result. However, the new function uses the new resource schema and it will run faster than the existing function.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,3 +110,6 @@ venv.bak/
 # vscode
 .vscode
 .idea/
+
+# Pre-commit
+.pre-commit-config.yaml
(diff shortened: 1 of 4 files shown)


norrishd (dismissed):
Hooray for finally getting all the segments!

norrishd (approved):
Hooray!

norrishd (approved):

PR #135, opened 2021-04-06
Update fitbit_activity.py

EwanNurse:
Changing the function called to access to diary study data as per bug raised by Freiburg group.


dkeden (approved):

dkeden:
All good once Buildkite passes!

PR #136, opened 2021-04-09
style: Change to use query variables

norrishd:
Alter the query to use GQL query variables instead of Python string formatting.

Two minor stylistic changes: using snake_case for variable names to comply with PEP8, and slightly simplify the pagination logic.


norrishd:
@matias-seer Just read the thread with Chee Wee and John where a different query was suggested - so maybe we need to further tweak the query and then do client-side filtering to get the channel group we need? Anyway up to you if you'd like to merge this into your branch :-)

norrishd:
Went through with Matias adding these directly to `SRR8_channel_Group_segments`

PR #138, opened 2021-04-14
Retrieve all segments when getting metadata

norrishd:
# Use pagination to retrieve all metadata segments
Use pagination to retrieve all segments when getting study metadata. NB that:
- `get_paginated_response()` can't be used here because pagination is required across items within a nested list, which is not currently supported by this function
- Switching to the resource schema (with its superior pagination abilities) proved problematic for a few reasons:
  - Several fields are either missing, null, or don't have permissions set in the analogous `resource` query
  - For some complex MySQL reason, on certain studies the analogous `resource` query can take >30 seconds (even with a small `limit`), meaning `seer-py` times out
  - Modifying the query to avoid timeout requires traversing the GQL graph in a weird order that involves a) lots of redundant data retrieval and b) client-side wrangling to get it into the expected `DataFrame` format

So all things considered, including how central the metadata DF is, and how we are currently blocked in Epiminder work because of this issue, we felt a little custom pagination for this query was the least evil (at least for now).

## Bonus
Tweak `get_channel_segments()` in a few ways:
- Return an empty `DataFrame` with all the expected columns when there are no channel group segments, instead of `None`. This is more consistent with other `seer-py` methods and probably easier to handle downstream
- Simplify the JSON-normalisation logic of the channel group column
- Rename all columns so they match the names of the columns returned by the various corresponding `get_all_metadata()` methods. This allows for easier concatenation.

### DIFF
--- a/seerpy/graphql.py
+++ b/seerpy/graphql.py
@@ -78,7 +78,7 @@ def get_string_from_list_of_dicts(list_of_dicts):
 
 
 GET_STUDY_WITH_DATA = """
-    query study_with_data($study_id: String!) {
+    query study_with_data($study_id: String!, $limit: PaginationAmount, $offset: Int) {
         study (id: $study_id) {
             id
             patient {
@@ -109,7 +109,7 @@ GET_STUDY_WITH_DATA = """
                 channelGroupType {
                     id
                 }
-                segments {
+                segments(limit: $limit, offset: $offset) {
                     id
                     startTime
                     duration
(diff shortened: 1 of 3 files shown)


norrishd:
@cheewee-seer No need to review/understand this; just letting you know that we've opted for pagination using the old schema for the time being, due to the challenges we discussed in switching to the `resource` schema.

matias-seer (approved):
looks good and works through testing on Minder

PR #139, opened 2021-05-08
Retry on 500 server error raised by gql

norrishd:
The algo has somewhat randomly started getting a lot of:
`Exception: {'statusCode': 500, 'errorCode': 'SERVER_ERROR', 'message': 'Internal server error'}`

These are raised by the `gql` library that `seer-py` uses internally and IDK, maybe they originate from our GQL server? Or further downstream as rate limits from S3? Opinions vary and I'm unclear. See also https://github.com/seermedical/seer-algo/issues/113.

In any case, pretty sure `seer-py` isn't catching these or retrying, so this is a tiny change to add this as an error to retry for. In the best case maybe this will largely fix things, and at the worst, will probably just drag things out for a few minutes before an instance turns into a zombie. Probably a worthwhile first step before taking a more drastic measure like switching to S3 direct?

P.s. my `pre-commit` YAPFed a bunch of bad indentation that had crept into the file, sorry that obscures the actual meaningful change.

### DIFF
--- a/seerpy/seerpy.py
+++ b/seerpy/seerpy.py
@@ -123,14 +123,18 @@ class SeerConnect:  # pylint: disable=too-many-public-methods
         See queries in graphql.py for structure of results returned
         """
         resolvable_api_errors = [
-            '502 Server Error', '503 Server Error', '504 Server Error', 'Read timed out.',
-            'NOT_AUTHENTICATED'
+            '502 Server Error',
+            '503 Server Error',
+            '504 Server Error',
+            'Read timed out.',
+            'NOT_AUTHENTICATED',
+            'SERVER_ERROR'  # 500 server error raised by gql library
         ]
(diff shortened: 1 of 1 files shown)


seerpy/seerpy.py
```diff
@@ -217,7 +221,6 @@ def get_paginated_response(self, query_string, variable_values, limit, object_pa
         offset = variable_values.get('offset', 0)  # Try get offset from variable values if exists
         result = []
         total_items_returned = 0
-        remaining_items = 0
```
norrishd, line 220: This is always redefined in the loop before being used, so isn't needed

ronrest:
Looks good to me

ronrest (approved):
looks good to me

matias-seer (approved):
looks good

PR #140, opened 2021-05-19
Add timeout param + switch to logging module

norrishd:
## Description
Three changes to make `seer-py` play nicer with `seer-algo`:

- Add an optional `timeout` param to the auth. We've been getting LOTS of timeouts with the default value of 30 seconds (as discovered by Chee Wee, manifesting as 460 errors in the ALB.) This has been leading to algo runs crashing and having to be re-tried, so hopefully this leads to a drastic reduction in that.
- Actually sleep for a few seconds if we get a 500/502 error trying to download from S3 (there was even a comment saying we should probably do this)
- Switch to the `logging` module instead of using `print` statements. This will also help with debugging in DataDog, determining which module errors arise in and the precise timings.

## Bonus fix
- In a number of methods that recently switched to internally use `get_paginated_response`, the `offset` argument has become orphaned  - this PR threads them back through as query variables.
- Add Ronny's handy `max_items` to another 1 or 2 methods where it slots in without issue.

## Style
I also got carried away and did some linting/docstring formatting.
- At some point we changed to a line length of 100 in YAPF, so update many docstrings to reflect this
- _seerpy.py_ is now over 2000 lines long (which seems Too long to me). So to help a tiny bit, delete param explanations for methods which are just thin wrappers around another method which is documented immediately above or below
- Conversely add in some missing params for non-duplicated functions
- Remove a couple of redundant parentheses
- Fix the odd typo

### DIFF
--- a/seerpy/auth.py
+++ b/seerpy/auth.py
@@ -7,28 +7,32 @@ from collections import namedtuple
 import datetime
 import getpass
 from glob import glob
-import os
 import json
+import logging
+import os
 import random
 import time
 
 import jwt
 import requests
 
+logger = logging.getLogger(__name__)
+
 KeyFileInfo = namedtuple('KeyFileInfo', ['key_path', 'key_id', 'region', 'default'])
 
 
 def get_auth(api_key_id=None, api_key_path=None, region=None, api_url=None, seer_auth=None,
-             use_email=None, email=None, password=None):
+             use_email=None, email=None, password=None, timeout=None):
     """
-    Get the correct Auth implemetation based on passed parameters and the existence of config files.
+    Get the correct Auth implementation based on passed parameters and the existence of config
+    files.
 
     If seer_auth is passed, it will be used and other parameters will be ignored.
     If use_email is True or email or password are passed, email authentication will be used.
     If no API key or id are passed and no API key files are found, email authentication will be
     used.
     If nothing is passed and API key files exist in ~/.seerpy matching the pattern seerpy*.pem then
-    API key authentiction will be used.
+    API key authentication will be used.
     API key files can contain the following information, separated by the '.' character:
     - a default indicator 'default' to indicate to use this file as a default
     - a region string 'au', 'de', 'uk', or 'us' indicating the region to use
@@ -56,8 +60,9 @@ def get_auth(api_key_id=None, api_key_path=None, region=None, api_url=None, seer
         The email address for a user's Seer account
     password : str, optional
         The password for a user's Seer account
+    timeout: int, optional
+            Timeout for queries made using the returned auth
     """
-
     if seer_auth:
         return seer_auth
 
@@ -68,9 +73,9 @@ def get_auth(api_key_id=None, api_key_path=None, region=None, api_url=None, seer
     if ((use_email is True)
             or (use_email is None and
                 (email or password or not (api_key_id or api_key_path or pem_files)))):
-        return SeerAuth(api_url, email, password)
+        return SeerAuth(api_url, email, password, timeout=timeout)
 
-    return SeerApiKeyAuth(api_key_id, api_key_path, region, api_url)
+    return SeerApiKeyAuth(api_key_id, api_key_path, region, api_url, timeout=timeout)
 
 
 # pylint: disable=no-self-use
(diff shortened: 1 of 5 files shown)


seerpy/auth.py
```diff
@@ -275,12 +280,13 @@ def _read_cookie(self):
 
 class SeerApiKeyAuth(BaseAuth):
     """
-    Creates an authenticated connection to the Seer API using an API key. This will become the
+    Create an authenticated connection to the Seer API using an API key. This is the recommended
```
norrishd, line 283: Seems like it's been long enough?

seerpy/auth.py
```diff
@@ -394,10 +402,11 @@ def _get_part_from_filename_parts(filename_parts, possible_values, part_title):
         if len(parts) > 1:
             raise ValueError(f'Multiple {part_title}s found in key file name')
         part = parts[0]
-        filename_parts = filename_parts.remove(part)
+        filename_parts.remove(part)
```
norrishd, line 405: Nothing is done with the return value, so no point reassigning it

seerpy/seerpy.py
```diff
@@ -53,15 +50,18 @@
 from . import utils
 from . import graphql
 
+logging.basicConfig(format='%(asctime)s %(message)s', level=logging.INFO)
```
norrishd, line 53: NB this is the same logger config used throughout our Python lambdas and `seer-algo`

seerpy/seerpy.py
```diff
@@ -141,12 +142,12 @@ def execute_query(self, query_string, party_id=None, invocations=0, variable_val
             return response
         except Exception as ex:
             if invocations > 4:
-                print('Too many failed query invocations. raising error')
+                logger.error('Too many failed query invocations, raising error')
                 raise
-            error_string = str(ex)
+            error_string = repr(ex)
```
norrishd, line 147: This way means we also get the error name, which can be helpful for debugging

seerpy/seerpy.py
```diff
@@ -292,14 +284,14 @@ def pandas_flatten(parent, parent_name, child_name):
         Example
         -------
         >>> df
-           start.id                           start.nested
-        0         A  [{'key1': 5, 'key2': 6}, {'key1': 7}]
-        1         B                          [{'key2': 8}]
+           top.id                              top.nested
```
norrishd, line 287: The example was wrong (I think my bad for way back)

seerpy/auth.py
```diff
@@ -320,6 +327,7 @@ def _get_parameters(self, api_key_id, api_key_path, region):
             raise ValueError('No API key file available')
 
         api_key_file = None
+        api_key_files = []
```
matias-seer, line 330: not that it is a problem, but why does this need to be initialised here?

matias-seer (dismissed):
LGTM

seerpy/seerpy.py
```diff
@@ -2031,14 +1942,10 @@ def get_study_ids_in_study_cohort(self, study_cohort_id, limit=200, offset=0, ma
...
         """
-        variable_values = {'study_cohort_id': study_cohort_id}
-        results = self.get_paginated_response(
-            graphql.GET_STUDY_IDS_IN_STUDY_COHORT_PAGED,
-            variable_values,
-            limit,
-            ['studyCohort', 'studies'],
-            max_items=max_items,
-        )
+        variable_values = {'study_cohort_id': study_cohort_id, 'offset': offset}
```
seerwill, line 1945: If this is fixing a bug should it be in its own PR?

seerpy/seerpy.py
```diff
@@ -2126,7 +2032,7 @@ def get_user_ids_in_user_cohort(self, user_cohort_id, limit=200, offset=0, max_i
         user_ids : list of str
             User IDs that are in the cohort
         """
-        variable_values = {'user_cohort_id': user_cohort_id}
+        variable_values = {'user_cohort_id': user_cohort_id, 'offset': offset}
```
seerwill, line 2035: As above - I feel like this should be in a separate PR if its addressing a bug.

seerwill (commented):
There are some unrelated changes attached i.e. offset changes, documentation, sleep etc which I think might have been better in a separate PR? Otherwise looks good to me.

There are some places where we use a custom `Auth`, however I can't spot any issues that might arise from these changes.

seerpy/seerpy.py
norrishd: Thanks for the feedback about the custom `Auth` @seerwill, my main worry was causing problems anywhere else by adding the timeout option.

I'll delete the `offset` fixes from this PR if you want them separate.

matias-seer (approved):
LGTM

PR #142, opened 2021-06-10
Fix dataChunk.end for last chunk in segment

timesler:
Before this change, the `SeerConnect.get_data_chunk_urls` method assumed that all chunks were the full `chunk_period`. This is typically not true for the last chunk in each segment. As a result, the `chunkData.end` column in the returned dataframe was incorrect for the last chunk in each segment. 

This change ensures that the end of the last chunk matches `segments.startTime + segments.Duration`.


norrishd (approved):
Nice pick up and speedy fix!

PR #149, opened 2021-10-11
FIX: only returning subset of labelgroups for study

ronrest:
The following two functions: 

- `client.get_label_groups_for_studies()`
- `client.get_label_groups_for_studies_dataframe()`

were only returning a subset of the label groups for each study (max of about 20). 
The query has been updated to fetch ALL label groups for each study.


seerpy/graphql.py
```diff
@@ -193,24 +193,28 @@ def get_string_from_list_of_dicts(list_of_dicts):
...
+    ) {
+        study(id: $study_id) {
+        id
+        name
+        labelGroups(limit: $limit, offset: $offset) {
+        id
+        name
+        description
+        labelType
+        numberOfLabels
```
norrishd, line 211: Nits, and no stress if you don't wanna change it since there's no linting enforceable for GQL queries when they're Python strings anyway, but it seems like convention in seer-py would be to indent the query variables, and indent again at `labelGroups.id`

seerpy/graphql.py
```diff
@@ -193,24 +193,28 @@ def get_string_from_list_of_dicts(list_of_dicts):
...
         }
+    }
     }"""
 
 
+
```
norrishd, line 217: Accidental newline?

seerpy/seerpy.py
```diff
@@ -934,43 +934,77 @@ def get_labels_string_dataframe(self, study_id, label_group_id, from_time=0, to_
...
+        """
+        results = self.get_paginated_response(graphql.GET_ALL_LABEL_GROUPS_FOR_STUDY_ID_PAGED,
+                                              variable_values=dict(study_id=study_id),
+                                              limit=limit,
+                                              object_path=["study"],
+                                              iteration_path=["labelGroups"],
+                                              )
+        return results
+
+    def get_label_groups_for_studies(self, study_ids, limit=50, labelgroup_limit=50):
```
norrishd, line 960: Is `limit` left in for backward compatibility or accidentally? In any case, since it's not doing anything I think it's a risky red herring to leave there. Could `labelgroup_limit` be removed and stick with `limit` to handle the paginating?

tests/test_seerpy.py
```diff
@@ -11,6 +11,9 @@
 from seerpy.seerpy import SeerConnect
 import seerpy.graphql as graphql
 
+from tests.test_data import label_groups_for_study
+from tests.test_data import label_groups_for_studies
```
norrishd, line 15: These could be combined into one line?

tests/test_seerpy.py
```diff
@@ -850,3 +847,28 @@ def test_empty(self, gql_client, unused_sleep, seer_connect):
 
         # check result
         assert result == expected_result
+
+
+@mock.patch('time.sleep', return_value=None)
+@mock.patch('seerpy.seerpy.GQLClient', autospec=True)
+class TestLabelGroups:
+    def test_get_label_groups_for_study(self, gql_client, unused_sleep, seer_connect):
```
norrishd, line 855: Noice testing!

tests/test_seerpy.py
```diff
@@ -850,3 +847,28 @@ def test_empty(self, gql_client, unused_sleep, seer_connect):
...
+        assert response == expected_seerpy_response
+
+    def test_get_label_groups_for_studies(self, gql_client, unused_sleep, seer_connect):
+        with mock.patch.object(seer_connect, "get_label_groups_for_study") as mock_stdy_labelgroups:
+            mock_stdy_labelgroups.side_effect = label_groups_for_studies.individual_study_responses
+            response = seer_connect.get_label_groups_for_studies(["study1","study2"])
+        assert label_groups_for_studies.expected_seerpy_response == response
+
+    def test_get_label_groups_for_studies_dataframe(self, gql_client, unused_sleep, seer_connect):
+        with mock.patch.object(seer_connect, "get_label_groups_for_study") as mock_stdy_labelgroups:
```
norrishd, line 870: Brendan once told me it's better for some reason to use the method decorator approach when you can equally well do so, i.e.

```python
@mock.patch.object(seer_connect, "get_label_groups_for_study")
def test_get_label_groups_for_studies(self, mock_stdy_labelgroups, gql_client, unused_sleep, seer_connect):
```

Having passed that on, I to this day don't know why it would matter, and am 100% happy for you to leave it like this if you like it more.

seerpy/graphql.py
ronrest: Fixed

seerpy/graphql.py
ronrest: removed

tests/test_seerpy.py
```diff
@@ -850,3 +847,28 @@ def test_empty(self, gql_client, unused_sleep, seer_connect):
...
+            mock_stdy_labelgroups.side_effect = label_groups_for_studies.individual_study_responses
+            response = seer_connect.get_label_groups_for_studies_dataframe(["study1","study2"])
+        assert label_groups_for_studies.expected_seerpy_df.equals(response)
+        print(response)
```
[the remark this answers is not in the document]
ronrest: Yep, accidentally forgot to remove it after debugging my tests.
Removed.

tests/test_seerpy.py
ronrest: I was going to change this to the way you suggested, but then realized that it made the function look a little uglier, so i have decided to leave it as it was (since functionally it is the same)

**Original:**

```python
def test_get_label_groups_for_studies(self, gql_client, unused_sleep, seer_connect):
    with mock.patch.object(seer_connect, "get_label_groups_for_study") as mock_stdy_labelgroups:
        mock_stdy_labelgroups.side_effect = label_groups_for_studies.individual_study_responses
        response = seer_connect.get_label_groups_for_studies(["study1","study2"])
    assert label_groups_for_studies.expected_seerpy_response == response
```

**If i change it:**

```python
@mock.patch.object(seer_connect, "get_label_groups_for_study")
def test_get_label_groups_for_studies(self, mock_stdy_labelgroups, gql_client, unused_sleep,
                                      seer_connect):
    mock_stdy_labelgroups.side_effect = label_groups_for_studies.individual_study_responses
    response = seer_connect.get_label_groups_for_studies(["study1","study2"])
    assert label_groups_for_studies.expected_seerpy_response == response
```

Similar reason for why I chose to call the variable `mock_stdy_labelgroups` instead of `mock_study_labelgroups`

seerpy/seerpy.py
ronrest: Fixed, so that `limit` applies at the label group level. And removed `labelgroup_limit`

ronrest:
ready for re-review

PR #150, opened 2021-10-21
Fix get_paginated_response when `iteration_path` returns empty on first iteration

ronrest:
There is a bug with `get_paginated_response()` such that if you are using an `iteration_path`, and the data at the `iteration_path` level is empty, on the first iteration, then it returns an empty list, instead of the top level information.

E.g. for a function that calls it, such as `client.get_label_groups_for_study("1234")`

it should return

```js
{'id': '1234',
 'name': 'someStudyName',
 'labelGroups': []}
```

But currently returns

```js
[]
```

Fixed it so it returns the correct thing.

### DIFF
--- a/tests/test_seerpy.py
+++ b/tests/test_seerpy.py
@@ -11,7 +11,13 @@ from seerpy import auth
 from seerpy.seerpy import SeerConnect
 import seerpy.graphql as graphql
 
-from tests.test_data import label_groups_for_study, label_groups_for_studies
+from tests.test_data import (
+    label_groups_for_study, 
+    label_groups_for_studies,
+    no_label_groups_for_study, 
+    no_label_groups_for_studies,
+    )
+
 
 # having a class is useful to allow patches to be shared across mutliple test functions, but then
 # pylint complains that the methods could be a function. this disables that warning.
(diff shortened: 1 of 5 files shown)


tests/test_seerpy.py
```diff
@@ -852,21 +858,43 @@ def test_empty(self, gql_client, unused_sleep, seer_connect):
...
     def test_get_label_groups_for_study(self, gql_client, unused_sleep, seer_connect):
+        # TEST WHEN STUDY CONTAINS TWO PAGES OF LABELGROUPS
         raw_paginated_responses = label_groups_for_study.raw_paginated_responses
         expected_seerpy_response = label_groups_for_study.expected_seerpy_response
+        gql_client.return_value.execute.side_effect = raw_paginated_responses
+        response = seer_connect.get_label_groups_for_study("study1")
+        assert response == expected_seerpy_response
 
+        # TEST WHEN STUDY CONTAINS NO LABELGROUPS
+        raw_paginated_responses = no_label_groups_for_study.raw_paginated_responses
```
norrishd, line 869: This seems like a good candidate for using `pytest.mark.parametrize()` to avoid duplicating code. Happy to point you to some examples or pair if you haven't used it before, lmk! But also all good by me just to ship this as is.

norrishd (approved):
Was this as painful a bug to figure out exists as I imagine it would've been? 😝 Nice going, I don't think I ever would've detected something like this!

tests/test_seerpy.py
norrishd: Yeah right on, feel free to merge as is, it's certainly better to have slightly verbose tests than none, and to have a bug fixed than not!

PR #151, opened 2022-01-24
Fix Pip Dependencies 

enjeyw:
Gql just got updated from 2.0.0 to 3.0.0. This is a breaking change, as it requires requests-toolbelt to be installed.
This means a current build of Seer-py right now won't work.

This PR adds the required deps, and pins all python deps to something sensible so that this is less likely to happen in the future.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@ gql==3.0.0
 requests==2.27.1
 numpy==1.22.1
 pandas==1.4.0
+argparse==1.4.0
 graphql-core==3.2.0
 requests-toolbelt==0.9.1
 pyjwt[crypto]==2.3.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,10 @@
 gql==3.0.0
-requests==2.27.1
-numpy==1.22.1
-pandas==1.4.0
-argparse==1.4.0
-graphql-core==3.2.0
+requests
+numpy
+pandas
+argparse
 requests-toolbelt==0.9.1
-pyjwt[crypto]==2.3.0
-tqdm==4.62.3
+graphql-core==3.2.0
+pyjwt[crypto]
+tqdm
 -e .
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,9 @@
-gql
-requests
-numpy
-pandas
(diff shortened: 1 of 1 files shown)


norrishd (dismissed):
Thanks for the fix!

requirements.txt
```diff
@@ -1,10 +1,10 @@
-gql
-requests
-numpy
-pandas
-# TODO: graphql-core 3.0.0 has a problem as at 02/12/2019
-# revisit in the future, we may be able to remove the restriction
-graphql-core<3.0.0
-pyjwt[crypto]
-tqdm
+gql==3.0.0
```
ronrest, line 1: if GQL is the only one that is causing the issue, then perhaps this is the only one whose version should be pinned. 
I am worried that pinning pandas and numpy will cause dependency version clashes with other repos that depend on seerpy, but which use a different version of pandas and numpy.

ronrest (changes requested):

requirements.txt
enjeyw: fair enough - will change!
(noting for the record that this leaves the door open for it to happen again)

norrishd: That's a good point to think about. Do you know which repos use seer-py? Off the top of my head I can think of:

- seer-worker
- seer-risk-algo
- seer-api-events (some lambdas)
- seer-disco (probably?)
- seer-holter (probably?)

ronrest: - I know `seer-holter` uses it. 
- I dont know what else.

enjeyw: just done a minimal pin required to pass tests :)
lemme know what u think @ronrest

ronrest: There used to be the following note on the requirements file:

```bash
# TODO: graphql-core 3.0.0 has a problem as at 02/12/2019
# revisit in the future, we may be able to remove the restriction
```
@norrishd , do you have any idea what the **problem** faced at that time was? And if it is covered by one of the tests? It has now been updated to `3.2.0`

norrishd: No I'm not sure, I think basically a bug in that version. I experimentally removed that pin in `pyql` and we've been using it for months now with no problem so presumably it's long been fixed.

ronrest: cool

ronrest (approved):

PR #152, opened 2022-01-25
Fix Setup

enjeyw:
Setup.py (required for installing seer-py as a pip dep) was missing `requests-toolbelt` as a requirement. This adds


ralphward (approved):

norrishd (approved):
Hmm I've never seen this before but seems fine if it fixes things!

PR #153, opened 2022-01-27
Simplify installation dependencies

norrishd:
Switch to installing the `gql` dependency using the recommended approach, as per [the docs](https://gql.readthedocs.io/en/latest/intro.html#installation).


enjeyw (approved):
Tested install locally and works for me!

